PR #6, opened 2019-09-19
Add comments to the exported elements

suleymanakbas91:
**Description**

- Add comments to the exported elements

**Related issue(s)**

See https://github.com/kyma-project/kyma/issues/5492

### DIFF
--- a/types/provider.go
+++ b/types/provider.go
@@ -1,23 +1,27 @@
 package types
 
-// Provider holds the details for the provider in use.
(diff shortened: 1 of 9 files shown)


types/provider.go
```diff
@@ -1,17 +1,23 @@
 package types
 
+// Provider holds the details for the provider in use.
```
bszwarc:
```suggestion
// Provider groups detailed information about the used provider. It stores information such as its type, the project it is used for, credentials and additional configurations. 
```

PR #12, opened 2019-09-25
Azure operator #5494

hisarbalik:


clebs:
After the complete overhaul of the hydroform terraform operator, this needs to be reimplemented in a different way.

Closing this PR in favor of #56

PR #27, opened 2019-10-24
update gardener template to match the new schema

obadrawi:


types/provider.go
```diff
@@ -14,14 +14,22 @@ type Provider struct {
...
 
 const (
-	// GCP stands for the Google Cloud Platform.
-	GCP ProviderType = "gcp"
-	// Azure stands for the Microsoft Azure Cloud Computing Platform.
-	Azure ProviderType = "azure"
-	// AWS stands for Amazon Web Services.
-	AWS ProviderType = "aws"
+	// GCPProvider stands for the Google Cloud Platform provider.
+	GCPProvider ProviderType = "gcp"
```
clebs: Is there a reason to change all these constants? We already defined the API like this for a reason, unless there is a need to change this, please revert it.

internal/operator/terraform.go
```diff
@@ -185,16 +200,16 @@ func (t *Terraform) newPlatform(providerType types.ProviderType, configuration m
 	var providerName string
 
 	switch providerType {
-	case types.GCP:
+	case types.GCPProvider:
```
clebs: Please completely revert this change.

types/provider.go
obadrawi: done..reverted

internal/operator/terraform.go
obadrawi: done

examples/gardener/main.go
```diff
@@ -13,10 +13,10 @@ import (
 )
 
 func main() {
-	projectName := flag.String("p", "", "Gardener project name")
+	projectName := flag.String("p", "garden-omar", "Gardener project name")
```
clebs: Please remove any personal or private data from the code.

examples/gardener/main.go
```diff
@@ -13,10 +13,10 @@ import (
...
 	machineType := flag.String("m", "n1-standard-4", "GCP machine type")
-	credentials := flag.String("c", "", "Path to the credentials file")
-	secret := flag.String("s", "", "Name of the secret to access the underlying provider of gardener")
+	credentials := flag.String("c", "C:\\Users\\D074188\\.kube\\config", "Path to the credentials file")
```
clebs: Please remove any personal or private data from the code.

examples/gardener/main.go
```diff
@@ -13,10 +13,10 @@ import (
...
+	secret := flag.String("s", "my-gcp", "Name of the secret to access the underlying provider of gardener")
```
clebs: Please remove any personal or private data from the code.

internal/gardener/gardener.go
```diff
@@ -89,7 +89,7 @@ func (g *gardenerProvisioner) Credentials(cluster *types.Cluster, provider *type
 		return nil, err
 	}
 
-	s, err := k8s.CoreV1().Secrets(fmt.Sprintf("garden-%s", provider.ProjectName)).Get(fmt.Sprintf("%s.kubeconfig", cluster.Name), metav1.GetOptions{})
+	s, err := k8s.CoreV1().Secrets(provider.ProjectName).Get(fmt.Sprintf("%s.kubeconfig", cluster.Name), metav1.GetOptions{})
```
clebs: The previous approach was already much better from a UX perspective, we can't assume that the users have to know they have to prefix their project name with "garden-"

examples/gardener/main.go
obadrawi: removed

examples/gardener/main.go
obadrawi: removed

examples/gardener/main.go
obadrawi: removed

internal/gardener/gardener.go
obadrawi: reverted back to old approach

internal/gardener/gardener.go
```diff
@@ -153,6 +153,16 @@ func (g *gardenerProvisioner) validate(cluster *types.Cluster, provider *types.P
 	} else {
 		errMessage += fmt.Sprintf(errs.CannotBeEmpty, "Provider.CustomConfigurations['target_provider']")
 	}
+	if v, ok := provider.CustomConfigurations["target_profile"]; ok {
```
clebs: Do we really need this configuration? As I can see, the possible profiles are: GCP, Azure and AWS. I guess they always match the `target_provider` (e.g. if `target_provider` is `gcp`, the `target_profile` will always be `gcpProfile`).
If this is the case, why do we need to explicitly provide the `target_profile` if we always can infer it from the `target_provider`?
Is there any case where `target_profile` and `target_provider` are not the same?

obadrawi: removed it from required variables and inferred the value from the provider

PR #53, opened 2019-12-10
Add configurable timeouts for operations

Szymongib:
**Description**

- Add configurable timeouts for operations

**Related issue(s)**


internal/gardener/gardener.go
```diff
@@ -24,6 +26,7 @@ const (
 
 type gardenerProvisioner struct {
 	operator operator.Operator
+	timeouts types.Timeouts
```
clebs: Options should be passed when calling Hydroform functions, not part of the provisioner.
See:
https://github.com/kyma-incubator/hydroform/blob/97160354ef3e83947e323214aeb8d93f57d2fd92/examples/gardener/gcp/main.go#L69

internal/gcp/gcp.go
```diff
@@ -18,6 +20,7 @@ import (
 // gcpProvisioner implements Provisioner
 type gcpProvisioner struct {
 	provisionOperator operator.Operator
+	timeouts          types.Timeouts
```
clebs: Same as for gardener, this should not be a field in the provisioner.

internal/config/timeouts.go
```diff
@@ -0,0 +1,31 @@
+package config
```
clebs: I would try to get rid of this package and just parse the timeouts option when loading the configuration. Also not make it part of the provisioner but instead adding it to the `Provider` type which already holds configurations for the cloud platform being targeted.

clebs:
Hi @Szymongib,

overall this idea is very good and it is a much needed feature for Hydroform.

Unfortunately, this implementation breaks the design this package follows.

See the comments I left on the points that should change and if you have any questions let me know!

internal/gardener/gardener.go
Szymongib: It is passed to the Hydroform function see here https://github.com/kyma-incubator/hydroform/pull/53/files#diff-defc0a0ee6263735b44fdcc9504d921bR38 only later the timeout are assigned to the provider in the `New` function here https://github.com/kyma-incubator/hydroform/pull/53/files#diff-dfa71d4200e54a500c09fed237cbbecdR49 but they come from options the same way as `Persistent()` and ` WithDataDir(dir string)`. Here is a full snippet:
```go
// New creates a new instance of gcpProvisioner.
func New(operatorType operator.Type, ops ...types.Option) *gcpProvisioner {
	// parse config
	os := &types.Options{}
	for _, o := range ops {
		o(os)
	}

	var op operator.Operator
	switch operatorType {
	case operator.TerraformOperator:
		tfOps := terraform_operator.ToTerraformOptions(os)
		op = terraform_operator.New(tfOps...)
	default:
		op = &operator.Unknown{}
	}

	return &gcpProvisioner{
		provisionOperator: op,
		timeouts:          os.Timeouts,
	}
}
```
And it can be called like that:
```go
	cluster, err := hf.Provision(cluster, provider, types.WithTimeouts(types.Timeouts{Create: 30*time.Minute}))
```

internal/config/timeouts.go
Szymongib: This is a helper function that is used during the `loadConfigurations` it is extracted to the separate package to not to duplicate logic. https://github.com/kyma-incubator/hydroform/pull/53/files#diff-dfa71d4200e54a500c09fed237cbbecdR230
I am open to changing the place of this function but I am not sure if it should be the configuration in the `Provider` as it is not really related to the provider but rather to underlying mechanisms of Hydroform (Terraform in this case).

clebs: The not duplicating makes sense definitely yes, and we might achieve it by pushing some of the parsing logic into the terraform operator. 
But the main goal of Hydroform is to abstract away the operator (terraform in this case).

Unfortunately the current implementation breaks the abstraction by adding those timeouts to the provisioners. We need to find a way to add this without leaking terraform specifics into hydroform.

Maybe we can move the load configuration common logic into the terraform operator and pass those timeouts as option to the API calls. That way the operator would be able to add those map entries if needed without disturbing the provisioners.

internal/gardener/gardener.go
clebs: Ah I see, then the approach on how the option is defined and given to hydroform is right.

The real issue is assigning it to the provisioners, which is already on our other thread.

kubeconfig.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
clebs: File committed by mistake, please remove it.

PR #69, opened 2020-01-13
enabled RBAC for azure always

a-thaler:
**Description**

- enabled kubernetes RBAC by default on azure
- ...
- ...

**Related issue(s)**
https://github.com/kyma-project/cli/issues/346


clebs:
Do we really need to enable RBAC? I usually do not enable it because when installing Kyma, its own access control is deployed into the cluster.

a-thaler:
yes, as all is based on the K8S RBAC

I tested the change manually and now RBAC was enabled

PR #79, opened 2020-02-04
Kind operator

tehcyx:
**Description**
First version of a kind operator.

- create/delete kind clusters with the terraform-provider that we have created [terraform-provider-kind](https://github.com/kyma-incubator/terraform-provider-kind)

What still is open (as on other providers):
- Find out how to read state from terraform state file
   - this actually blocks receiving a properly filled kubeconfig, file can be obtained still via `kind` cli
- Figure out if the terraform provider supports all use cases we'd need to setup kind clusters (for testing)
   - A custom kind config would be next, to increase number of nodes (https://github.com/kyma-incubator/terraform-provider-kind/issues/6)


provision/examples/README.md
```diff
@@ -7,4 +7,5 @@ The following examples show you how to use Hydroform for cluster provisioning:
 * [GCP](../examples/gcp/README.md)
 * [Gardener/GCP](../examples/gardener/gcp/README.md)
 * [Gardener/Azure](../examples/gardener/azure/README.md)
-* [Gardener/AKS](../examples/gardener/aws/README.md)
\ No newline at end of file
+* [Gardener/AKS](../examples/gardener/aws/README.md)
```
clebs: I know this was not edited in this PR, but it should be `Gardener/AWS` instead of `Gardener/AKS`.

tehcyx: done

provision/examples/kind/README.md
```diff
@@ -0,0 +1,31 @@
+# Provision a Kind cluster
+
+## Overview
+
+This example shows you how to use Hydroform to provision a Kind cluster.
+
```
bszwarc:
```suggestion
## Prerequisites
To provision a Kind cluster you need Docker with memory settings configured to support the cluster you want to run.
```
Let's just add a prerequisites section here instead of `Configure Kind` later. I think it makes more sense to put it like this.

provision/examples/kind/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Installation
+
+### Configure Kind
+
+To provision an Kind cluster you need:
+
+1. Docker installed on the machine
+
+2. Docker configured to have the memory settings to support the cluster you want to run
+
```
bszwarc:
```suggestion
```
Remove these lines -> content moved to `Prerequisites`

tehcyx:
> Added comments.

Changes applied @bszwarc

PR #86, opened 2020-02-12
Update terraform provider

rakesh-garimella:
**Description**

- Update terraform

**Related issue(s)**

kyma-incubator/terraform-provider-gardener#34

### DIFF
--- a/provision/examples/gardener/azure/main.go
+++ b/provision/examples/gardener/azure/main.go
@@ -51,7 +51,7 @@ func main() {
 			"networking_pods":        "100.96.0.0/11",
 			"networking_services":    "100.64.0.0/13",
 			"networking_type":        "calico",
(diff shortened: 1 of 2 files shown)


provision/examples/gardener/azure/main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 			"networking_pods":        "100.96.0.0/11",
 			"networking_services":    "100.64.0.0/13",
 			"networking_type":        "calico",
-			"service_endpoints":      []string{"microsoft.test", "microsoft2.test"},
+			"service_endpoints":      []string{""},
```
suleymanakbas91: why do we need to pass this? can't we just drop service_endpoints?

rakesh-garimella: we need a valid service endpoints (like Azure Storage etc) to work else cluster creation fails. So thats why I have made it empty string for the program to function properly.

Issue #97, opened 2020-02-27
Don't use readAll to read files


clebs:
**Description**
Same as in the Kyma CLI on https://github.com/kyma-project/cli/issues/363, we are using `ioutil.ReadAll()` to read a file on hydroform.

Please replace with `io.Copy()` the same way as done on the CLI.

PR #98, opened 2020-02-27
Using copy instead of read all

shorim:
**Description**

- replacing ioutil.ReadAll() with io.Copy() as suggested here https://haisum.github.io/2017/09/11/golang-ioutil-readall/

**Related issue(s)**
#97

### DIFF
--- a/install/example/example.go
+++ b/install/example/example.go
@@ -1,11 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"flag"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"log"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


install/example/example.go
```diff
@@ -105,12 +106,12 @@ func fetchFile(url string) (string, error) {
 		return "", fmt.Errorf("received non OK status code while getting a file from url: %s", url)
 	}
 
-	data, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
+	var data bytes.Buffer
+	if _, err := io.Copy(&data, resp.Body); err != nil {
```
Szymongib: Is it really necessary in the example? It is not accessed concurrently and we load it whole into the memory anyways so I do not really see any benefits. As `ioutil.ReadAll(resp.Body)` is simpler I think it would be more clear in the example.

shorim: Even if it is not accessed concurrently, it might happen that the resp.Body is huge.
So, I think it is better to be safe using io.Copy(), as it will only uses fixed 32KB buffer to copy from source to destination until EOF

Issue #104, opened 2020-04-02
Gardener AWS: Incorrect "zone" property is used


clebs:
**Description**
The gardener AWS terraform template is using a variable `aws_zone` to set the zone which is never set, since hydroform sets the `zone` variable for all providers in gardener.
We need to use the zone variable aslo for AWS like the others.

**Expected result**
When setting the `zone` flag the Gardener AWS cluster uses that value in provisioning

**Actual result**
Validation is passed sucessfully but later terraform prompts to set the `aws_zone` variable and provisioning can't be done.

clebs:
While working on this issue, gardener updated its validation, making a bug visible that we had until now in Azure.
This PR fixes it in the gardener terraform provider: https://github.com/kyma-incubator/terraform-provider-gardener/pull/73 and it will be integrated together with the AWS zone improvements into hydroform and the CLI

PR #106, opened 2020-04-06
Refactor zones and CIDR management in gardener

clebs:
**Description**

- Zones are now handled properly for all gardener providers.
- CIDR networks are configured properly for all gardener providers.
- Renamed "zone" variable to be consistent.
- Made networking properties optional (now calculated and generated based on `vnetcidr` and `workercidr`).
- Updated all gardener examples.

**Related issue(s)**
#104


provision/examples/gardener/azure/main.go
```diff
@@ -40,19 +41,16 @@ func main() {
...
 			"worker_minimum":         2,
 			"machine_image_name":     "coreos",
 			"machine_image_version":  "2303.3.0",
-			"networking_nodes":       "10.250.0.0/19",
-			"networking_pods":        "100.96.0.0/11",
-			"networking_services":    "100.64.0.0/13",
 			"networking_type":        "calico",
 			"service_endpoints":      []string{""},
-			"zone":                   []string{"1"},
+			"zones":                  []string{},
```
rakesh-garimella, line 53: why do you remove the zone from here ?

clebs: Because otherwise you have to specify `networking_pods` for the gardener validation to happen.

The interesting thing is if I do not specify it, gardener assigns the same value to it on the server side, so I guess the server side validation has a bug.

This way I just keep the example more simple.

clebs: Mystery solved, please see comment in https://github.com/kyma-incubator/hydroform/issues/104

PR #117, opened 2020-06-29
K8s 118: Apply the CRDs before applying the CR

rakesh-garimella:
**Description**

- With k8s 1.18 we cannot apply CRDs and CR together in the same file. So change the behavior to apply first the CRD and then the CR

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/8360


install/installation/installation.go
```diff
@@ -310,46 +322,53 @@ func (k KymaInstaller) installTiller(tillerYaml string, createFunction func([]k8
...
 
-	k8sInstallerObjects, err := k8s.ParseYamlToK8sObjects(k.decoder, installerYaml)
+	k8sInstallerObjects, err := k8s.ParseYamlToK8sObjects(k.decoder, yamlFile)
 	if err != nil {
-		return fmt.Errorf("failed to parse Installer yaml file to Kubernetes dynamicClientObjects: %w", err)
+		return fmt.Errorf("failed to parse yaml file to Kubernetes dynamicClientObjects: %w", err)
 	}
 
-	var installationCR *v1alpha1.Installation
-	installationCR, k8sInstallerObjects, err = k.extractInstallationCR(k8sInstallerObjects)
```
Szymongib, line 330: We do extract the Installation CR from the artifacts here so I do not understand what this change actually achieves.

install/installation/installation.go
```diff
@@ -310,46 +322,53 @@ func (k KymaInstaller) installTiller(tillerYaml string, createFunction func([]k8
...
+			k.infof("Applying Installation CR modifications...")
+			k.installationOptions.installationCRModificationFunc(installationCR)
+			k.infof("Applying Installation CR...")
+			err = k.applyInstallationCR(installationCR)
+			if err != nil {
+				return fmt.Errorf("failed to apply Installation resources: %w", err)
+			}
+			k.infof("Installer CR deployed.")
 
-	_, err = createResourcesFunc(k8sInstallerObjects)
```
Szymongib, line 340: Here we apply all resources from installation artifacts besides Installation CR so the Installation CRD should be applied here.

install/installation/installation.go
```diff
@@ -310,46 +322,53 @@ func (k KymaInstaller) installTiller(tillerYaml string, createFunction func([]k8
...
+		if err != nil {
+			return fmt.Errorf("failed to apply Installer resources: %w", err)
+		}
+		k.infof("Installer CRD deployed.")
 	}
 
-	k.infof("Applying Installation CR modifications...")
-	k.installationOptions.installationCRModificationFunc(installationCR)
-	k.infof("Applying Installation CR...")
-	err = k.applyInstallationCR(installationCR)
```
Szymongib, line 348: The Installation CR is applied here after the rest of the artifacts so I do not see where the problem lies exactly.

install/installation/installation.go
```diff
@@ -74,6 +74,8 @@ type Installation struct {
 	TillerYaml string
 	// InstallerYaml is a content of yaml file with all resources related to and required by Installer
 	InstallerYaml string
+	// InstallerCRYaml is a content of yaml file with the list of the packages being installed
+	InstallerCRYaml string
```
Szymongib, line 78: Currently, Installation CR is a part of InstallerYaml and this API should be aligned with it. If we want to make such a change, then the Installation CR should be moved out of InstallrYaml artifact.

rakesh-garimella:
Related PR for CLI: https://github.com/kyma-project/cli/pull/485
Related PR on Prow: https://github.com/kyma-project/test-infra/pull/2517

install/installation/installation.go
rakesh-garimella: By default on master the installation CR is not a part of the InstallerYaml. Currently during the release process we are merging the two. This PR: https://github.com/kyma-project/test-infra/pull/2517 will change the behavior so that we get two files on release page similar to the one in master.

rakesh-garimella:
@Szymongib I had to remove the extraction part since i did not wanted to send merged yaml file anymore from the cli. I have moved that part in to `deployInstaller` function definition.

install/installation/installation.go
```diff
@@ -221,6 +223,11 @@ func (k KymaInstaller) PrepareInstallation(artifacts Installation) error {
 		return err
 	}
 
+	err = k.deployInstallerForIstallation(artifacts.InstallerCRYaml)
+	if err != nil {
+		return err
+	}
+
```
Szymongib, line 230: I do not think it is best to use the same function for applying Installer Yaml as well as Installation CR. Also, Installation CR artifact may not be provided (in case of older versions) so we need to handle both cases.
I would suggest doing something like that:
```go
	if artifacts.InstallerCRYaml != "" {
		err := k.deployInstallationCRYaml(artifacts.InstallerCRYaml)
		if err != nil {
			return err
		}
	}
```
And creating two new functions for deploying installation CR - from Yaml and from Go struct - that could look something like that:
```go
func (k KymaInstaller) deployInstallationCRYaml(yamlFile string) error {

	var installationCR v1alpha1.Installation
	err := yaml.NewDecoder(bytes.NewBuffer([]byte(yamlFile))).Decode(&installationCR)
	if err != nil {
		return fmt.Errorf("error decoding installation CR: %s", err.Error())
	}

	return k.deployInstallationCR(&installationCR)
}

func (k KymaInstaller) deployInstallationCR(installationCR *v1alpha1.Installation) error {

	_, found := installationCR.Labels[installationActionLabel]
	if found {
		delete(installationCR.Labels, installationActionLabel)
	}

	k.infof("Applying Installation CR modifications...")
	k.installationOptions.installationCRModificationFunc(installationCR)
	k.infof("Applying Installation CR...")
	err := k.applyInstallationCR(installationCR)
	if err != nil {
		return fmt.Errorf("failed to apply Installation resources: %w", err)
	}
	k.infof("Installer CR deployed.")
	return nil
}
```

Then we can simplify the  `deployInstaller` function, but still, use the extraction functionality for older versions:
```go
func (k KymaInstaller) deployInstaller(installerYaml string, createResourcesFunc func(resources []k8s.K8sObject) ([]*unstructured.Unstructured, error)) error {
	k.infof("Deploying Installer...")

	k8sInstallerObjects, err := k8s.ParseYamlToK8sObjects(k.decoder, installerYaml)
	if err != nil {
		return fmt.Errorf("failed to parse Installer yaml file to Kubernetes dynamicClientObjects: %w", err)
	}

	// If Installation CR found in installerYaml - apply it otherwise do nothing
	var installationCR *v1alpha1.Installation
	installationCR, k8sInstallerObjects, err = k.extractInstallationCR(k8sInstallerObjects)
	if err != nil {
		if err != errorInstallationCRNotFound {
			return fmt.Errorf("failed to get Installation CR: %w", err)
		}
	}

	_, err = createResourcesFunc(k8sInstallerObjects)
	if err != nil {
		return fmt.Errorf("failed to apply Installer resources: %w", err)
	}

	if installationCR != nil {
		return k.deployInstallationCR(installationCR)
	}

	return nil
}
```

install/installation/installation.go
```diff
@@ -246,6 +253,11 @@ func (k KymaInstaller) PrepareUpgrade(artifacts Installation) error {
 		return err
 	}
 
+	err = k.deployInstallerForUpgrade(artifacts.InstallerCRYaml)
+	if err != nil {
+		return err
+	}
+
```
Szymongib, line 260: I suggest handling that the same way as described above.

install/installation/installation.go
```diff
@@ -397,23 +416,6 @@ func (k KymaInstaller) applyInstallationCR(installationCR *v1alpha1.Installation
...
-			}
-
-			k8sObjects = append(k8sObjects[:i], k8sObjects[i+1:]...)
-			return installationCR, k8sObjects, nil
-		}
-	}
-
-	return nil, k8sObjects, fmt.Errorf("installation object not found in the objects slice")
-}
-
```
Szymongib, line 416: To properly handle artifacts with and without the Installation CR Yaml we could add an error that would provide distinction between when something went wrong and when Installation CR is not present if Yaml file.
We can do it as follows:
```go
var (
	errorInstallationCRNotFound = fmt.Errorf("error installationCR not found in objects")
)

// extractInstallationCR finds and removes first Installation CR from the slice of K8sObjects and returns it
func (k KymaInstaller) extractInstallationCR(k8sObjects []k8s.K8sObject) (*v1alpha1.Installation, []k8s.K8sObject, error) {
	for i, k8sObject := range k8sObjects {
		if k8sObject.GVK.Group == v1alpha1.Group && k8sObject.GVK.Kind == "Installation" {
			installationCR, ok := k8sObject.Object.(*v1alpha1.Installation)
			if !ok {
				return nil, nil, fmt.Errorf("unexpected type of Installation object: %T, failed to cast to *Installation", k8sObject.Object)
			}

			k8sObjects = append(k8sObjects[:i], k8sObjects[i+1:]...)
			return installationCR, k8sObjects, nil
		}
	}

	return nil, k8sObjects, errorInstallationCRNotFound
}
```
or use some additional bool variable.

PR #118, opened 2020-07-13
Make hydroform work with the latest kind provider

tehcyx:
**Description**

This PR enables cluster creation via hydroform, correctly reading the kubeconfig from the created kind cluster. *DO NOT MERGE* unless there is a release for the kind provider created or it won't work.
After this is merged work that I started on the CLI can continue so that this will work on provisioning kind clusters with the CLI as well.

- Correctly read kubeconfig from created cluster
- Write back kubeconfig for use in CLI

**Related issue(s)**
Using all the changes from latest kind provider: https://github.com/kyma-incubator/terraform-provider-kind/pull/12

### DIFF
--- a/provision/internal/operator/terraform/provider.go
+++ b/provision/internal/operator/terraform/provider.go
@@ -15,14 +15,26 @@ import (
 // TODO remove this file when the gardener provider is on the official terraform registry
 
 const (
-	providerURL     = "https://github.com/kyma-incubator/terraform-provider-gardener/releases/download/%s/terraform-provider-gardener-%s-%s"
-	providerName    = "terraform-provider-gardener"
-	providerVersion = "v0.0.9"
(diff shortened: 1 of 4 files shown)


tehcyx:
I've just released the kind provider: https://github.com/kyma-incubator/terraform-provider-kind/releases/tag/v0.0.1

provision/internal/operator/terraform/provider.go
```diff
@@ -15,14 +15,26 @@ import (
...
 )
 
 // initGardenerProvider will check if the gardener provider is available and download it if not.
 
 func initGardenerProvider() error {
+	return initProvider(gardenerProviderName, gardenerProviderVersion, gardenerProviderURL)
+}
+func initKindProvider() error {
+	return nil
+	// return initProvider(kindProviderName, kindProviderVersion, kindProviderURL)
```
clebs, line 34: Shouldn't this be uncommented? Right now this provider will not work unless I manually download the binary previously.

tehcyx: indeed, done.

PR #143, opened 2020-11-25
Improve parallel-installation logging

colunira:
**Description**

- Pass logging function as a configurable parameter
- Add log prefix to internal packages

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/9347

### DIFF
--- a/parallel-install/example/example.go
+++ b/parallel-install/example/example.go
@@ -57,7 +57,7 @@ func main() {
 		HelmTimeoutSeconds:            60 * 8,
 		BackoffInitialIntervalSeconds: 3,
 		BackoffMaxElapsedTimeSeconds:  60 * 5,
-		Log: 						   log.Infof,
+		Log:                           log.Infof,
 	}
 
 	installer, err := installation.NewInstallation(prerequisitesContent,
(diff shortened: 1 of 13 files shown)


parallel-install/example/example.go
```diff
@@ -2,8 +2,9 @@ package main
 
 import (
 	"flag"
+	log "github.com/sirupsen/logrus"
```
Tomasz-Smelcerz-SAP, line 5: Do we need logrus here? Perhaps built-in log package is enough?

parallel-install/pkg/installation/installation.go
```diff
@@ -51,13 +50,17 @@ func NewInstallation(prerequisites [][]string, componentsYaml string, overridesY
 	}, nil
 }
 
+func (i *Installation) SetupLogger(log func(string, ...interface{})) {
```
Tomasz-Smelcerz-SAP, line 53: This function should be defined as top-level public function in the config package.

parallel-install/example/example.go
colunira: I used it for testing purposes, I can remove it now :)

Tomasz-Smelcerz-SAP (approved):

PR #145, opened 2020-11-25
Context cancel() support in exponent-backoff retries

Tomasz-Smelcerz-SAP:
**Description**

Add support for Context in exponential-backoff retries in helm client wrapper
This may look as we're testing the external library (which has its own tests), but in fact we're testing our usage of it and if it behaves as expected.

- Configure backoff library with our main context
- Tests for common scenarios

**Related issue(s)**
See also https://github.com/kyma-project/kyma/issues/9347


strekm (approved):
/meow

PR #147, opened 2020-11-26
Add prerequisites timeout to parallel install library

colunira:
**Description**

- Move namespace creation to prerequisites
- Add prerequisites timeout
- Move some logic to separate functions
- Add installation package tests

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/9347

### DIFF
--- a/parallel-install/example/example.go
+++ b/parallel-install/example/example.go
@@ -2,7 +2,11 @@ package main
 
 import (
 	"flag"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/components"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/engine"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/overrides"
 	"io/ioutil"
+	"k8s.io/client-go/kubernetes"
 	"log"
 
 	"os"
@@ -71,14 +75,30 @@ func main() {
 
 	config.SetupLogger(log.Printf)
 
-	err = installer.StartKymaInstallation(restConfig)
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		//return fmt.Errorf("Unable to create internal client. Error: %v", err)
+	}
+
+	overridesProvider, err := overrides.New(kubeClient, installer.OverridesYamls, installer.Cfg.Log)
+	if err != nil {
+		//return fmt.Errorf("Unable to create overrides provider. Error: %v", err)
+	}
+
+	prerequisitesProvider := components.NewPrerequisitesProvider(overridesProvider, installer.ResourcesPath, installer.Prerequisites, installer.Cfg)
+	componentsProvider := components.NewComponentsProvider(overridesProvider, installer.ResourcesPath, installer.ComponentsYaml, installer.Cfg)
+
+	engineCfg := engine.Config{WorkersCount: installer.Cfg.WorkersCount}
+	eng := engine.NewEngine(overridesProvider, componentsProvider, installer.ResourcesPath, engineCfg)
+
+	err = installer.StartKymaInstallation(*prerequisitesProvider, overridesProvider, eng)
 	if err != nil {
 		log.Printf("Failed to install Kyma: %v", err)
 	} else {
 		log.Println("Kyma installed!")
 	}
 
-	err = installer.StartKymaUninstallation(restConfig)
+	err = installer.StartKymaUninstallation(*prerequisitesProvider, overridesProvider, eng)
 	if err != nil {
 		log.Fatalf("Failed to uninstall Kyma: %v", err)
 	}
--- a/parallel-install/example/example.go
+++ b/parallel-install/example/example.go
@@ -56,8 +56,8 @@ func main() {
 
 	installationCfg := config.Config{
 		WorkersCount:                  4,
-		CancelTimeout:                 60 * time.Second,
-		QuitTimeout:                   1 * time.Second,
+		CancelTimeout:                 20 * time.Minute,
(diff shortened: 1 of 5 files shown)


parallel-install/pkg/installation/installation_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"time"
+)
+
+const cancelTimeout = 50
+const quitTimeout = 150
+const cancelTimeoutMillisecond = time.Duration(50) * time.Millisecond
+const quitTimeoutMillisecond = time.Duration(150) * time.Millisecond
+
+
+func Test_CancelTimeout(t *testing.T) {
```
Tomasz-Smelcerz-SAP, line 20: Please provide a tests for entire flow (StartKymaInstallation)
- happy path
- cancel timeout in prerequisites (ensure normal install doesn't start)
- quit timeout in the middle of prerequisites installation (ensure it quits immediately even if some prerequisites are still being installed in the background)

parallel-install/pkg/installation/installation_test.go
```diff
@@ -0,0 +1,93 @@
...
+	hc := &mockHelmClient{
+		cancelWithTimeout: true,
+	}
+	component := newComponent(hc)
+
+	//when
+	err := i.installPrerequisites(ctx, cancelFunc, []components.Component{component}, cancelTimeoutMillisecond, quitTimeoutMillisecond)
+
+	//then
+	assert.Error(t, err, "Kyma installation failed due to the timeout")
```
Tomasz-Smelcerz-SAP, line 35: It would be nice here to show that if we have e.g: 5 components and there's the timeout during installation of the third, the fourth and fifth one are skipped. We'd have test coverage for `prerequisites` package "for free" then.

parallel-install/example/example.go
```diff
@@ -71,14 +75,32 @@ func main() {
...
 	config.SetupLogger(log.Printf)
 
-	err = installer.StartKymaInstallation(restConfig)
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		log.Printf("Failed to create kube client. Exiting...")
+		os.Exit(1)
+	}
+
+	overridesProvider, err := overrides.New(kubeClient, installer.OverridesYamls, installer.Cfg.Log)
```
suleymanakbas91, line 84: All this logic should be internal to the library. The consumer of the library should not care for overrides provider, prerequisite provider, engine creation, etc. The consumer should be able to create an installation object with the configuration and then call the installation and uninstallation functions.

suleymanakbas91 (changes requested):

parallel-install/example/example.go
colunira: I'm afraid that if we move this back we will not be able to test it at all. I understand your concerns and I don't like this solution, too, but after we have tests we can refactor this code to make it more user-friendly and still be sure that everything works as expected.

Tomasz-Smelcerz-SAP: @suleymanakbas91 I can agree with the exception of kubeClient - this should be provided by the user, I don't see how it should be library responsibility to create a k8s client. It should be an external dependency.

parallel-install/pkg/installation/installation_test.go
```diff
@@ -0,0 +1,352 @@
...
+			overridesProvider := &mockOverridesProvider{}
+			eng := engine.NewEngine(overridesProvider, provider, "", engine.Config{
+				WorkersCount: 2,
+				Log:          log.Printf,
+			})
+
+			err := i.StartKymaInstallation(kubeClient, provider, overridesProvider, eng)
+
+			assert.Error(t, err)
+			assert.EqualError(t, err, "Kyma prerequisites installation failed due to the timeout")
```
Tomasz-Smelcerz-SAP, line 62: Please also add time-based assertions with a comment as in "due to quit timeout" test case

parallel-install/pkg/installation/installation_test.go
```diff
@@ -0,0 +1,352 @@
...
+			overridesProvider := &mockOverridesProvider{}
+			eng := engine.NewEngine(overridesProvider, provider, "", engine.Config{
+				WorkersCount: 2,
+				Log:          log.Printf,
+			})
+
+			err := i.StartKymaInstallation(kubeClient, provider, overridesProvider, eng)
+
+			assert.Error(t, err)
+			assert.EqualError(t, err, "Kyma installation failed due to the timeout")
```
Tomasz-Smelcerz-SAP, line 115: Please also add time-based assertions with a comment as in "due to quit timeout" test case

parallel-install/pkg/installation/installation_test.go
```diff
@@ -0,0 +1,352 @@
...
+			start := time.Now()
+			err := inst.StartKymaInstallation(kubeClient, provider, overridesProvider, eng)
+			end := time.Now()
+
+			elapsed := end.Sub(start)
+
+			assert.Error(t, err)
+			assert.EqualError(t, err, "Force quit: Kyma installation failed due to the timeout")
+
+			// Prerequisites and two components installation lasts 300 ms
```
Tomasz-Smelcerz-SAP, line 146: 3 prereq + 2 components * 71[ms] = 355[ms]. The comment mentions 300[ms]. Am I missing something?

colunira: I am using 2 workers, so components are installed/uninstalled in parallel. I will add this as a comment, as it might be confusing in the future.

Tomasz-Smelcerz-SAP (approved):

suleymanakbas91 (approved):

PR #164, opened 2020-12-29
Clean logs for alpha deploy cmd

tobiscr:
**Description**

Add support for clean log output (works with async-deploy cmd) and add verbose flag.

**Related issue(s)**


suleymanakbas91 (approved):
Looks good in general. My only concern is not being able to see the error messages on non-verbose mode.

tobiscr:
> Looks good in general. My only concern is not being able to see the error messages on non-verbose mode.

Thanks @suleymanakbas91 for bringing this up.

@strekm : Think we should discuss this point to enable the CLI to provide more detailed error messages.

strekm (approved):

parallel-install/README.md
```diff
@@ -31,7 +32,7 @@ See all available configuration options for the `config.Config` type:
 | HelmTimeoutSeconds | `int` | `360` | Timeout for the underlying Helm client. |
 | BackoffInitialIntervalSeconds | `int` | `1` | Initial interval used for exponential backoff retry policy. |
 | BackoffMaxElapsedTimeSeconds | `int` | `30` | Maximum time used for exponential backoff retry policy. |
-| Log | `func(format string, v ...interface{})` | `fmt.Printf` | Function used for logging. |
+| Log | `func(format string, v ...interface{})` | `fmt.Printf` | Function used for logging. To modify the logging behavior, set a custom logging function. E.g. for disabling any log-output, just provide an empty logging function implementation (`func(f string, v ...interface{}){}`) |
```
kazydek, line 35:
```suggestion
| Log | `func(format string, v ...interface{})` | `fmt.Printf` | Function used for logging. To modify the logging behavior, set a custom logging function. For example, to disable any log output, provide an empty logging function implementation (`func(f string, v ...interface{}){}`). |
```

kazydek (approved):

Issue #160, opened 2020-12-22
parallel-installation lib metadata implementation


strekm:
**Description**

The installation library should accept kyma metadata and preserve it on a cluster after successful kyma deployment.

**Reasons**

AC:
- deployment metadata is preserved on a cluster after kyma deployment.
- it is possible to query deployment metadata
- metadata contains kyma version

**Attachments**

strekm:
No strong contract on the structure, having key-value structure. Data to store:
- profile: evaluation | production
- version (e.g. 1.17, PR-1234,3464564, local)
- status: Started, Deployed, Failed
- reason: e.g error message, failed components

PR #165, opened 2020-12-31
Parallel install metadata

strekm:
**Description**

- metadata in parallel-installation lib

**Related issue(s)**
see #160

### DIFF
--- a/parallel-install/README.md
+++ b/parallel-install/README.md
@@ -41,6 +41,7 @@ Once you have a configured `Deployment` instance, use the following functions ac
 
 - `StartKymaDeployment` - Starts the deployment process. First, prerequisites are deployed linearly. Then, the components' deployment continues in parallel.
 - `StartKymaUninstallation` - Starts the uninstallation process. The library uninstalls the components first, then it proceeds with the prerequisites' uninstallation in reverse order.
+- `ReadKymaMetadata` - Retrieves Kyma metadata, e.g. version.
 
 ### Example
 
--- a/parallel-install/README.md
+++ b/parallel-install/README.md
@@ -41,7 +41,7 @@ Once you have a configured `Deployment` instance, use the following functions ac
 
 - `StartKymaDeployment` - Starts the deployment process. First, prerequisites are deployed linearly. Then, the components' deployment continues in parallel.
 - `StartKymaUninstallation` - Starts the uninstallation process. The library uninstalls the components first, then it proceeds with the prerequisites' uninstallation in reverse order.
-- `ReadKymaMetadata` - Retrieves Kyma metadata, e.g. version.
+- `ReadKymaMetadata` - Retrieves Kyma metadata, such as Kyma version.
 
 ### Example
 
--- a/parallel-install/pkg/deployment/deployment.go
+++ b/parallel-install/pkg/deployment/deployment.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/metadata"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/components"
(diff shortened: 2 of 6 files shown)


parallel-install/pkg/deployment/deployment.go
```diff
@@ -134,6 +136,10 @@ func (i *Deployment) StartKymaUninstallation() error {
 	return nil
 }
```
colunira, line 138: Please add a godoc comment like above (e.g. `//StartKymaDeployment implements the Installer.StartKymaDeployment contract.`)

parallel-install/pkg/metadata/metadata.go
```diff
@@ -110,17 +135,17 @@ func (p *Provider) WriteKymaDeployed() error {
...
+	return retryOperation(func() error {
+		return p.writeKymaMetadata(meta)
+	})
 }
 
 func (p *Provider) writeKymaMetadata(data *KymaMetadata) error {
 	cmData := metadataToCM(data)
 
-	//TODO retries
 	kymaMetadataCM, err := p.kubeClient.CoreV1().ConfigMaps(metadataNamespace).Get(context.TODO(), metadataName, metav1.GetOptions{})
```
colunira, line 146: I think you forgot about the retries here?

parallel-install/pkg/metadata/metadata.go
```diff
@@ -138,7 +163,6 @@ func (p *Provider) writeKymaMetadata(data *KymaMetadata) error {
 		return err
 	}
 
-	//TODO update CM
 	kymaMetadataCM.Data = cmData
 	_, err = p.kubeClient.CoreV1().ConfigMaps(metadataNamespace).Update(context.TODO(), kymaMetadataCM, metav1.UpdateOptions{})
```
colunira, line 167: Should we put retries here as well?

parallel-install/pkg/metadata/metadata.go
```diff
@@ -14,6 +16,9 @@ var metadataName = "kyma"
 //TODO it's in default ns cause at starting installation kyma-installer ns doesn't exist, this needs to be changed and metadata should be written to protected ns
 var metadataNamespace = "default"
 
+var initialInterval = time.Duration(3) * time.Second
+var maxElapsedTime = time.Duration(30) * time.Second
```
colunira, line 20: Have you tested this values? How many retries do we get with this configuration? Maybe retries shouldn't be exponential in this case.

colunira (changes requested):

parallel-install/pkg/metadata/metadata.go
strekm: i implemented retries on usages of this method, so the whole method is retried.

parallel-install/pkg/metadata/metadata.go
strekm: as my above comment

parallel-install/pkg/metadata/metadata.go
strekm: adjusted values

colunira (approved):

parallel-install/pkg/deployment/deployment.go
```diff
@@ -53,43 +63,95 @@ type Installer interface {
 //See overrides.New for details about the overrides contract.
 //
 //resourcesPath is a local filesystem path where components' charts are located.
-func NewDeployment(prerequisites [][]string, componentsYaml string, overridesYamls []string, resourcesPath string, cfg config.Config, processUpdates chan<- ProcessUpdate) (*Deployment, error) {
+func NewDeployment(prerequisites [][]string, componentsYaml string, overridesYamls []string, resourcesPath string, cfg config.Config, processUpdates chan<- ProcessUpdate, kubeClient kubernetes.Interface) (*Deployment, error) {
```
tobiscr, line 66: Would recommend to move this to the config.Config struct. The method signature is already very long. Please see #166 for the refactored signature.

Maybe worth to think about moving parts which are coupled to the deployment-package / simple parameters to the config object (e.g. also the componentListPath, Overries). This would have the benefit to add parameter-validation to the config-object to identify missing parameters.

parallel-install/pkg/deployment/deployment.go
```diff
@@ -27,20 +33,24 @@ type Deployment struct {
 	ResourcesPath string
 	Cfg           config.Config
 	// Used to send progress events of a running install/uninstall process
-	ProcessUpdates chan<- ProcessUpdate
+	ProcessUpdates   chan<- ProcessUpdate
+	kubeClient       kubernetes.Interface
+	metadataProvider metadata.MetadataProvider
 }
 
 type Installer interface {
```
tobiscr, line 41: Do we really need that interface any longer as we have only one implementation of it?

strekm: we agreed to get rid of this interface

parallel-install/pkg/deployment/deployment.go
strekm: we agreed to leave it like that for now and deal with it in following PR

parallel-install/README.md
```diff
@@ -41,6 +41,7 @@ Once you have a configured `Deployment` instance, use the following functions ac
 
 - `StartKymaDeployment` - Starts the deployment process. First, prerequisites are deployed linearly. Then, the components' deployment continues in parallel.
 - `StartKymaUninstallation` - Starts the uninstallation process. The library uninstalls the components first, then it proceeds with the prerequisites' uninstallation in reverse order.
+- `ReadKymaMetadata` - Retrieves Kyma metadata, e.g. version.
```
klaudiagrz, line 44:
```suggestion
- `ReadKymaMetadata` - Retrieves Kyma metadata, such as Kyma version.
```

tobiscr (approved):

klaudiagrz (approved):

Issue #169, opened 2021-01-14
parallel-installation lib logging refactoring


strekm:
**Description**

Use logging API instead of passing log function around.

AC:
- log function is not used anymore

**Reasons**

Having clean, consistent API across the library.

tobiscr:
Closing this issue as PR was merged

PR #179, opened 2021-01-26
Refactor logging in parallel-install lib

mjakobczyk:
**Description**

- remove old log function
- introduce zap logger dependency
- create logger factory function that should be used across the library
- remove one global logger passed in `Config` to let every struct has it's own logger
- add `Verbose` flag to Config to easy logger toggling
- replace all Go SDK log calls with zap logger calls

**Related issue(s)**

Implements #169


strekm (changes requested):
I'm not a big fan of such direct dependency to zap logger (imagine zap replaced with another logger), I would expect a tiny interface in logger.go and this interface used around our stuff

tobiscr (approved):

tobiscr:
> I'm not a big fan of such direct dependency to zap logger (imagine zap replaced with another logger), I would expect a tiny interface in logger.go and this interface used around our stuff

@strekm : Maybe worth to bring this topic up in the quality sig-group to come up with one logging abstractoin-layer used globally in Kyma which uses internally ZAP. Otherwise we have a decouple logging layer in Hydroform, another one in CLI etc. etc.

strekm (approved):

parallel-install/pkg/components/component.go
```diff
@@ -39,37 +40,37 @@ type KymaComponent struct {
 	//OverridesGetter is a function that returns overrides for the release.
 	OverridesGetter func() map[string]interface{}
 	HelmClient      helm.ClientInterface
-	Log             func(format string, v ...interface{})
+	Log             *zap.SugaredLogger
```
Tomasz-Smelcerz-SAP, line 43: The Idea behind logging in the library was the logging should be generic. In other words, the library should be usable by any "client" (some other software providing the "main" function") and that the library should not impose any particular logger - it's the client's responsibility to choose one.
I see this hard-coded zap library dependency as a violation of this idea. What's the profit of such approach compared to the previous contract?

parallel-install/pkg/components/prerequisites.go
```diff
@@ -29,7 +31,7 @@ func NewPrerequisitesProvider(overridesProvider overrides.OverridesProvider, res
 		HelmTimeoutSeconds:            cfg.HelmTimeoutSeconds,
 		BackoffInitialIntervalSeconds: cfg.BackoffInitialIntervalSeconds,
 		BackoffMaxElapsedTimeSeconds:  cfg.BackoffMaxElapsedTimeSeconds,
-		Log:                           cfg.Log,
+		Log:                           logger.NewLogger(cfg.Verbose),
```
Tomasz-Smelcerz-SAP, line 34: I don't think lower-level packages should actually care about logging level. Logger should be configured somewhere higher in the stack, and the components should be given ready-to-use logger. What's the value of propagating information about "Verbosity" across all layers?

parallel-install/pkg/components/prerequisites.go
```diff
@@ -38,7 +40,7 @@ func NewPrerequisitesProvider(overridesProvider overrides.OverridesProvider, res
 		resourcesPath:     resourcesPath,
 		components:        components,
 		helmConfig:        helmCfg,
-		log:               cfg.Log,
+		log:               logger.NewLogger(cfg.Verbose),
```
Tomasz-Smelcerz-SAP, line 43: Previous version was actually shorter and simpler. What's the benefit here?

parallel-install/pkg/deployment/deployment.go
```diff
@@ -229,7 +232,7 @@ func (i *Deployment) deployPrerequisites(ctx context.Context, cancelFunc context
 		Context:       ctx,
 		KubeClient:    i.kubeClient,
 		Prerequisites: p,
-		Log:           i.cfg.Log,
+		Log:           logger.NewLogger(i.cfg.Verbose),
```
Tomasz-Smelcerz-SAP, line 235: Do we really need to instantiate a separate logger for every component? Is that actually safe? See: https://github.com/uber-go/zap/issues/99

parallel-install/example/example.go
```diff
@@ -4,7 +4,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"log"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
+	"go.uber.org/zap"
```
Tomasz-Smelcerz-SAP, line 8: Is "zap" a good choice? See: [Note that zap only supports the two most recent minor versions of Go.](https://github.com/uber-go/zap)

parallel-install/pkg/logger/logger.go
```diff
@@ -0,0 +1,37 @@
+package logger
+
+import (
+	"go.uber.org/zap"
```
Tomasz-Smelcerz-SAP, line 4: Again, this makes our library hard-coded to a single logger. What if the client has other preferences? Well, now there are two loggers in the project. What if CLI decides they need some other logger? We'd have to make another PR touching 20 files or so. Take a look at how Helm is solving that - they're using an abstract log function. The current logging is based on the same idea.
Personally, I see this as a step backward in the usability of the library.

parallel-install/example/example.go
tobiscr: Please bring that point to the logging-working group during the SIG quality group. We should not discuss it in this PR review :)

parallel-install/pkg/components/component.go
tobiscr: AGree.. the logger API should not couple the application to a parituclar logger implemenation.

I'm anyway wodering why Golang has not a common logging-abstraction API in place yet.. we should focus on establishing something like this: https://github.com/birkirb/loggers

parallel-install/pkg/components/prerequisites.go
mjakobczyk: Verbosity flag was required to instantiate separate logger for every component. That's the way it was discussed to be done. After followup, we have decided to change the strategy.

parallel-install/pkg/components/prerequisites.go
mjakobczyk: It was discussed to create a separate logger for every component. After followup, we have decided to change the strategy.

parallel-install/pkg/deployment/deployment.go
mjakobczyk: After followup, we have decided to change the strategy. Now logger will be created once on the highest level and passed around.

parallel-install/example/example.go
mjakobczyk: It was picked as the suggestion to be used came from PoC.

parallel-install/pkg/logger/logger.go
mjakobczyk: After followup, we have decided to change the strategy. Now logger will be created once on the highest level and passed around.

Tomasz-Smelcerz-SAP (approved):

PR #182, opened 2021-02-02
Overrides Conflict Resolution Strategy

ralikio:
**Description**

Introduce `merge` or `replace` flags to allow differ between algorithms used for override resources merge.

- Merging/replacing configuration overrides added by config maps and secrets.

**Related issue(s)**

#350


install/merger/merger.go
```diff
@@ -0,0 +1,33 @@
+package merger
+
+const (
+	OnConflictLabel   string = "on-conflict"
+	ReplaceOnConflict string = "replace"
+)
+
+type Data interface {
```
akgalwas, line 8: This is purely design related remark. This is used in one of the following ways:
- User calls Update
- User calls Merge and then Update

Usually when I see that the interface contains a methods that must be called is a specific order I try to understand if the interface could be redesigned.

We could have the following:
```
type Data interface {
  Labels() *map[string]string
  Override() error // makes Update 
  Merge() error // makes merge and then Update
}
```

The implementation would call either `Override` or `Merge`.  You can ignore it. I still think that this abstraction is not suitable here.

install/merger/merger_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func NewData() *DataImpl {
+	return &DataImpl{
+		labels: map[string]string{
+			"label": "value",
+		},
+	}
+}
+
+func (d *DataImpl) Labels() *map[string]string {
```
akgalwas, line 22: In such a case I would rather use mock than fake. Mocks have nice `AssertCalled()` and `AssertNotCalled()` methods.

install/merger/types/config_map_client.go
```diff
@@ -0,0 +1,12 @@
+package types
```
akgalwas, line 1: In general, in Go we place all related types within one file. It should rather be in configmap_override.go.

install/merger/types/configmap_override_test.go
```diff
@@ -0,0 +1,99 @@
+package types
+
+import (
+	"context"
+	"github.com/kyma-incubator/hydroform/install/merger/types/mocks"
```
akgalwas, line 5: `ConfigMapClient` mock is not checked in.

install/merger/types/configmap_override_test.go
```diff
@@ -0,0 +1,99 @@
...
+				Labels: map[string]string{
+					"anotherKey": "anotherValue",
+				},
+			},
+		},
+		Client: client,
+	}
+}
+
+type MockError struct {
```
akgalwas, line 94: This type is redundant. You can use `errors.New()` from `errors` package.

install/merger/types/config_map_client.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ConfigMapClient interface {
```
akgalwas, line 9: Useful trick for interfaces. You could annotate the interface to be able to use go generate:
`//go:generate mockery -name= ConfigMapClient`

install/merger/types/configmap_override.go
```diff
@@ -0,0 +1,31 @@
+package types
+
+import (
+	"context"
+	. "github.com/kyma-incubator/hydroform/install/util"
```
akgalwas, line 5: I think using dot in file import is not recommended in this case. Please see: https://github.com/golang/go/wiki/CodeReviewComments#import-dot

install/merger/types/secret_override_test.go
```diff
@@ -0,0 +1,88 @@
+package types
+
+import (
+	"context"
```
akgalwas, line 4: SecretClient mock is not checked in.

akgalwas (changes requested):

install/k8s/generic_client.go
```diff
@@ -87,25 +89,14 @@ func (c GenericClient) ApplyConfigMaps(configMaps []*corev1.ConfigMap, namespace
 	return nil
 }
 
-func (c GenericClient) updateConfigMap(client corev1Client.ConfigMapInterface, cm *corev1.ConfigMap) error {
```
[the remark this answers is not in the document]
ralikio: > I'm not sure that the approach is suitable for this case. Frankly I would prefer to add the logic to both `updateConfigMap` and `updateSecret`. You may ask : "_hey, this is a duplication_". Yes, it is. Rob Pike once said : "_A little copying is better than a little dependency._". I will paraphrase that: "_A little copying is better than a little abstraction_".

It seems reasonable but the definitions are blurry. The fact is that the amount of code duplication in the Hydroform component is significant and introduces some side effects. For example, when you look at all copied code from kyma-incubator/hydroform/install/installation/configuration.go we might want start analysing the code until we reach the methods newOverridesConfigMap and newOverridesSecret. I asked myself that if any method was copied in regard to secret/configMap distinction then why did these two methods differ? To understand I have to go all the way to CLI component or just give back the investigation. This is what a troublesome dependency sounds like to me. Such issues are enough to make at least little changes one by one in controlled (especially tested) manner.

> Software Engineering is about making choices. Similarity in `updateConfigMap` and `updateSecret` is obvious. I don't like it. But I would think twice before trying to remove the duplication due to limited facilities for generic programming in Go. 

It is not certain now but I hope that if generics are introduced in future releases of GO then the PR code could be simplified even more. It is a fundament for the future. I agree, the code reflects only small logic changes but does it mean that we should immediately forget about good practices?

> In this solution we added 357 lines of code (along with tests) in eight files containing three interfaces and two structures. Correct me if I'm wrong, but the only goal was to remove duplication in one file.

As you pointed out the point of extraction was to remove duplication, but we cannot forget that additionally we get good code separation, easier tests and no possibility that there are some unintentional changes introduced only in one fragment of code and not in its copies.  

Bringing the metric of code lines does not make sens. Is there a number that is not acceptable, like 300 is ok but 357 is not? The actual logic takes 34 lines in merger.go, 34x2 for configmap/secret_override.go (I have removed unnecessary Name() method when writing the response), 2x13 in _client.go and 2x5 for code application in generic_client.go. The rest contains tests and mocks (I did not commit mocks previously generated with mockery). The main advantage of such separation is that the logic (I admit - small) is well separated and tested. What gives even bigger number of lines is the fact that I like to keep them short. We are living in the era of ultra-wide monitors but when I have multiple windows open (for example when doing diff in IDE) it is easier to read such code than to deal with "infinite" horizontal scrolling. Finally, I tend to separate abstractions across multiple files which increases the number of extra imports. 

> In my opinion this is over-engineering.
> Such an abstraction would make sense if:
> 
> * there would be a possibility to introduce new implementations (e.g. some new entity for storing configuration similar to ConfigMap or Secret) ; in such a case there will be more duplication and developer would need to remember to implement override strategy for new type

Of course, because of the code being nicely separated from the rest we have that possibility to apply it to additional domain abstractions and extend it if needed. Will it happen in the future? I do not know. But the idea of writing stuff with extensibility in mind and concern separation is something that I aim to accomplish. It is hard to foresee the future, the only thing that is certain that the code changes and we should introduce measures for that code. 

> * the duplication would span many different files

I am not sure that thinking in files is correct. It would be better to think in packages because the files in packages can be rearranged freely. I have introduced only one package (merger) because I want to centralise my changes wherever possible.

> * the code needed for abstracting things up would be only slightly longer than "the ugly" solution

It is longer but not as much as pointed previously. The fact that I wrote multiple test files is just for the sake of the solution being well checked. What is more the solution is centralised and without unnecessary duplications. I agree that I wrote duplicated client.go classes but that is the minimum that has to be copied without generics. 

> 
> I'm open to discussion, but in my opinion neither is the case. When coding in Go (well, and any other language) we should strive toward simplicity. Abstraction has cost and the cost shouldn't overweight the gain.

Testing

install/merger/types/config_map_client.go
ralikio: I did not commit generated classes for Client interfaces. Now I added additional mock for merger's Data.

install/merger/merger.go
ralikio: Such approach would force on us further interface changes if modifying merger logic. I see interfaces as a separation between implementation and its usages. If add Merger/Override in the interface we are effectively breaking encapsulation because now is the external components (not the merger) that decide when and how to do merger's logic.

install/k8s/generic_client.go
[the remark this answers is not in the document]
akgalwas: Well, what about `ApplyConfigMaps` and `ApplySecrets`?
There is duplication.

```
func (c GenericClient) ApplyConfigMaps(configMaps []*corev1.ConfigMap, namespace string) error {
	client := c.coreClient.ConfigMaps(namespace)

	for _, cm := range configMaps {
		_, err := client.Create(context.Background(), cm, v1.CreateOptions{})
		if err != nil {
			if k8serrors.IsAlreadyExists(err) {
				err = c.updateConfigMap(client, cm)
				if err != nil {
					return fmt.Errorf("config map %s already exists, failed to updated config map: %s", cm.Name, err.Error())
				}
				continue
			}
			return fmt.Errorf("failed to apply %s config map: %s", cm.Name, err.Error())
		}
	}
	return nil
}


```
func (c GenericClient) ApplySecrets(secrets []*corev1.Secret, namespace string) error {
	client := c.coreClient.Secrets(namespace)

	for _, sec := range secrets {
		_, err := client.Create(context.Background(), sec, v1.CreateOptions{})
		if err != nil {
			if k8serrors.IsAlreadyExists(err) {
				err = c.updateSecret(client, sec)
				if err != nil {
					return fmt.Errorf("secret %s already exists, failed to updated secret: %s", sec.Name, err.Error())
				}
				continue
			}
			return fmt.Errorf("failed to apply %s secret: %s", sec.Name, err.Error())
		}
	}

	return nil
}

```
Why we don't fix that? Why duplication in `updateConfigMap` and `updateSecret` needed some additional abstraction? Do you plan to refactor it as well?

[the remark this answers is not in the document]
Comment (akgalwas) [reply install/k8s/generic_client.go:90]: Please, see my response below.

> It seems reasonable but the definitions are blurry. The fact is that the amount of code duplication in the Hydroform component is significant and introduces some side effects. For example, when you look at all copied code from kyma-incubator/hydroform/install/installation/configuration.go we might want start analysing the code until we reach the methods newOverridesConfigMap and newOverridesSecret. I asked myself that if any method was copied in regard to secret/configMap distinction then why did these two methods differ? To understand I have to go all the way to CLI component or just give back the investigation. This is what a troublesome dependency sounds like to me. Such issues are enough to make at least little changes one by one in controlled (especially tested) manner.
I don't get this arguments. Duplication is significant in all Go projects. 

> It is not certain now but I hope that if generics are introduced in future releases of GO then the PR code could be simplified even more. It is a fundament for the future. I agree, the code reflects only small logic changes but does it mean that we should immediately forget about good practices?
 
Well, what is crucial about good practices is that they always have good reasoning behind it. Let's take DRY. This is pretty obvious what we gain by applying this principle.  On the other hand, there are cases when we decide to not follow this principle. Let's take microservice world. We decide to have duplicated code in different services in case introducing some library with common logic would break one of the most important microservice feature: ability to deploy independently. So we sacrifice the DRY principle for the sake of having a service we can release independently. 
This long story has one purpose. Some 'good practices' don't apply in each and every software domain - what works for monoliths not necessarily will for microservices. Something what is considered a good practice in one language is not in other. 

I'm not in favour of following 'good practices' in a religious manner. What should be done from engineering perspective is considering pros and cons.  We always need to compare possible solutions, understand the cost and the value. 

> Of course, because of the code being nicely separated from the rest we have that possibility to apply it to additional domain abstractions and extend it if needed. 

But the fundamental question is whether we need it? I can't think of any use case. The whole point about abstraction in OO is flexibility. Let's take some more concrete example. In Application Gateway there is an abstraction layer handling authentication.  It is plausible that we would need to add some new authentication method. From time perspective I could say that the abstraction payed off. But we don't need such flexibility in this case. Here is a great article on over-abstracting things: https://www.quantcast.com/blog/death-by-1000-layers-the-perils-of-over-abstraction-in-java/

Some general remarks. Adding additional degree of flexibility to the design is not always needed. If I would have to design **fork** I would ask myself the following question: "what is the use case".  **Folding fork** is good when you have limited space in your backpack, and in general it have very specific use case: outdoor activity. But you probably won't use it for a Sunday lunch.  So adding the feature of folding adds flexibility but is not always needed by the end user.    

> Will it happen in the future? I do not know. But the idea of writing stuff with extensibility in mind and concern separation is something that I aim to accomplish. It is hard to foresee the future, the only thing that is certain that the code changes and we should introduce measures for that code.

No one knows, but adding some new abstractions 'just in case' is not a good idea. 

> I am not sure that thinking in files is correct. It would be better to think in packages because the files in packages can be rearranged freely. I have introduced only one package (merger) because I want to centralise my changes wherever possible.

My point was that having duplication in many files is worse than having it in one file. As it comes to the merger package I don't think it has correct place in the directory structure. It should be under k8s as this is a detail of interacting with k8s. 

> It is longer but not as much as pointed previously. The fact that I wrote multiple test files is just for the sake of the solution being well checked. What is more the solution is centralised and without unnecessary duplications. I agree that I wrote duplicated client.go classes but that is the minimum that has to be copied without generics.

It is significantly longer. We could add _ifs_ in two functions and modify generic_client_test.go. I agree, it is not beautiful. But adding a new package containing a new (unneeded) abstraction with three interfaces and two structures is too much. 

> Bringing the metric of code lines does not make sens. Is there a number that is not acceptable, like 300 is ok but 357 is not? The actual logic takes 34 lines in merger.go, 34x2 for configmap/secret_override.go (I have removed unnecessary Name() method when writing the response), 2x13 in _client.go and 2x5 for code application in generic_client.go. The rest contains tests and mocks (I did not commit mocks previously generated with mockery).
 
You don't get my point. This is not about exact number. 300 lines is either wrong. 
We could do one of the following things:
- Write _if_ in two places and some tests in generic_client_test.go
- Add an extra abstraction that increases cognitive load and spend some time writing several hundreds lines of code.

In my opinion it is better to apply "ugly" solution as there is no elegant way to get rid of the duplication. My sense of aesthetics suffers when I see duplication in Go. But in this word it is a norm unfortunately. In Go we use code generation that implicates duplication as there is no generics mechanism. I'm not pretty certain if Go 2.0 will fix that. In C++ or Rust I would use templating mechanism that would remove duplication in elegant manner - with few lines of code without introducing any new interfaces.

 Moreover with your solution you cannot hide the duplication. You remove it in one place and it shows up in another. Please, compare `config_map_override_test.go` and  `secret_override_test.go` they are almost identical, the only difference is the type they operate on. 
  

> What gives even bigger number of lines is the fact that I like to keep them short. We are living in the era of ultra-wide monitors but when I have multiple windows open (for example when doing diff in IDE) it is easier to read such code than to deal with "infinite" horizontal scrolling. Finally, I tend to separate abstractions across multiple files which increases the number of extra imports. 

Well, the generic_client.go file has about 200 lines.  This is not "infinitive". 

To sum it up: the solution is over-engineered.

File: install/merger/merger.go (hunk 1 above)
Comment (akgalwas) [reply install/merger/merger.go:8]: I don't get your point. Creating interfaces with some hidden assumption about  method calling order is not a good practice IMHO.

**Comment by ralikio** [comment]
I do not see that there is any room for discussion. I do not agree with the proposed solution. However, respecting your experience I applied the changes requested by you. I hope you would find the code more to YOUR likings now.

**Comment by ralikio** [comment]
I do not see that there is any room for discussion. I do not agree with the proposed solution. However, respecting your experience I applied the changes requested by you. I hope you would find the code more to your likings now.

**Comment by ralikio** [comment]
I do not agree with the proposed solution. However, respecting your experience I applied the changes requested by you. I hope you would find the code more to your likings now.

**Review by akgalwas — APPROVED** [review]

**Comment by PK85** [comment]
@ralikio please join kyma-incubator organisation, and contact me on priv

File: install/k8s/generic_client.go [hunk 10]
```diff
@@ -12,16 +11,23 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
+	. "k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
```
koala7659, line 15: Please remove this "dot" import.

install/k8s/generic_client.go
```diff
@@ -12,16 +11,23 @@ import (
...
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	. "k8s.io/client-go/kubernetes/typed/core/v1"
 	corev1Client "k8s.io/client-go/kubernetes/typed/core/v1"
```
koala7659, line 23: Please remove this "dot" import.

install/installation/installation.go
```diff
@@ -317,10 +317,10 @@ func (k KymaInstaller) StartInstallation(context context.Context) (<-chan Instal
 	return stateChannel, errorChannel, nil
 }
 
-func (k KymaInstaller) CheckInstallationState() (InstallationState, error) {
+func (k KymaInstaller) CheckInstallationState(installationClient installationTyped.InstallationInterface) (InstallationState, error) {
```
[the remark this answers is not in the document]
ralikio: I have discussed the change with @Maladie. There was a problem that broke old interface that made it impossible to upgrade Hydroform dependency in control-plane. We yet have to discuss the required changes with the team responsible for CLI.

install/installation/installation.go
```diff
@@ -317,10 +317,10 @@ func (k KymaInstaller) StartInstallation(context context.Context) (<-chan Instal
...
 	return checkInstallationState(k.installationClient)
 }
-func (k KymaInstaller) TriggerUninstall() error {
+func (k KymaInstaller) TriggerUninstall(installationClient installationTyped.InstallationInterface) error {
```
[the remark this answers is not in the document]
ralikio: The same as above.

install/merger/merger.go
ralikio: Interfaces removed.

install/installation/installation.go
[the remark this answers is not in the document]
ralikio: I have discussed the change with @Maladie. There was a problem that broke old interface that made it impossible to upgrade Hydroform dependency in control-plane. We yet have to discuss the required changes.

PR #183, opened 2021-02-02
New installation: Split deployment and deletion

tobiscr:
**Description**

Refactoring applied to split deployment and deletion to improve config-parameter processing.

**Related issue(s)**


parallel-install/example/example.go
```diff
@@ -80,27 +80,33 @@ func main() {
...
 	}
 
 	log.Printf("Kyma version: %s", kymaMeta.Version)
 	log.Printf("Kyma status: %s", kymaMeta.Status)
 
-	err = installer.StartKymaUninstallation()
+	//Delete Kyma
+	deleter, err := deployment.NewDeletion(installationCfg, overrides, kubeClient, progressCh)
+	if err != nil {
+		log.Fatalf("Failed to create installer: %v", err)
```
tobiscr, line 107:
```suggestion
		log.Fatalf("Failed to create deleter: %v", err)
```

parallel-install/pkg/deployment/deployment_test.go
```diff
@@ -240,7 +240,7 @@ func TestDeployment_StartKymaDeployment(t *testing.T) {
 func TestDeployment_StartKymaUninstallation(t *testing.T) {
```
suleymanakbas91, line 240: Wouldn't it be better if we move deletion related tests to another file, i.e. deletion_test.go?

parallel-install/pkg/config/config.go
```diff
@@ -41,14 +41,22 @@ type Config struct {
...
 	}
 	if err := c.pathExists(c.ComponentsListFile, "Components list"); err != nil {
 		return err
 	}
+	return nil
+}
+
+// ValidateDeployment verifies that deployment specific options are provided
+func (c *Config) ValidateDeployment() error {
+	if err := c.ValidateDeletion(); err != nil { //deployment requires all core options
```
suleymanakbas91, line 57: What if someone writes a deletion specific logic in `ValidateDeletion` and forgets that deployment validation also depends on that? Wouldn't it be better to have a base function that both uses for validating the common things?

suleymanakbas91 (changes requested):

parallel-install/pkg/deployment/deployment_test.go
tobiscr: You got me! Was too lazy here.. will do that :)

parallel-install/pkg/config/config.go
tobiscr: MAybe worth to directly introduce a DeploymentConfig and DeletionConfig object which have decoupled functions

tobiscr: Tried to split the configuration into a Deployment- and Deletion specific entity, but the refactoring would become quite huge as the config-object is passed between so many instances. Will track this as improvement for later.

tobiscr: The single test function is now split into multiple test functions, one for each public config validation function.

suleymanakbas91 (approved):

PR #185, opened 2021-02-04
Remove incompatible components on k3d cluster

tobiscr:
**Description**

Some components cannot be deployed on a locally running K3d cluster. As these components are deprecated and not required to run Kyma locally, they will be automatically removed from the component list if the cluster is based on K3d.

**Related issue(s)**


strekm (approved):
/meow

PR #186, opened 2021-02-04
Add library that allows running functions locally

pPrecel:
**Description**

- Support operations on docker containers (`pkg/docker` and `pkg/docker/runtimes`)

**Related issue(s)**

https://github.com/kyma-project/cli/pull/717
https://github.com/kyma-project/kyma/issues/9508


pPrecel:
TODO: 
- [ ] missing tests

m00g3n:
So far it looks great (however still checking it yet).

A couple of comments: 
- We could add image pull before `run`. Right now the image needs to be pulled independently.
- For `inline functions` we could use the config file's `sourcePath` property to determine the context path instead `$PWD`,when the `-d` flag is not passed.
- For `git type functions` we could assume that the context is located in the same directory as the config file instead $PWD

pPrecel:
> So far it looks great (however still checking it yet).
> 
> A couple of comments:
> 
> * We could add image pull before `run`. Right now the image needs to be pulled independently.
> * For `inline functions` we could use the config file's `sourcePath` property to determine the context path instead `$PWD`,when the `-d` flag is not passed.
> * For `git type functions` we could assume that the context is located in the same directory as the config file instead $PWD

Thanks! It's done, but I'm still testing new solution.

m00g3n (approved):
LGTM

PR #192, opened 2021-02-11
Fix bug ignoring return error when installation fails

clebs:
**Description**

- Parallel install was not returning errors when components would fail installing. Making jobs run tests even when installation was not successful: https://status.build.kyma-project.io/?job=kyma-integration-gardener-gcp-parallel


strekm (approved):
/meow

Issue #177, opened 2021-01-21
parallel-installation CRDs installation improvement


strekm:
**Description**

At the moment we have a job in the cluster-essentials component, which is applying CRDs. Move this logic directly to the library and implement retries.

AC:
- CRDs installed via k83 client
- retries implemented
- if pod-preset is still present in the cluster-essentials install only this sub-chart

**Reasons**

Having an easy and stable way of installing CRDs

PR #205, opened 2021-02-19
Introduce Preinstaller in installation lib

mjakobczyk:
**Description**

- rewrite logic from cluster-essentials crd-install-job for installation lib
- cover installation of CRDs
- cover creation of Namespaces

**Related issue(s)**

Implements #177


parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
+package preinstaller
```
Tomasz-Smelcerz-SAP, line 1: It's good to describe the package here unless you want a separate go.doc file. Take a look at other packages for examples.

parallel-install/pkg/preinstaller/output.go
```diff
@@ -0,0 +1,14 @@
+package preinstaller
```
Tomasz-Smelcerz-SAP, line 1: These types are used only by things in "preinstaller.go". I think it makes sense to move them there.

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
...
+		applier:       applier,
+		cfg:           cfg,
+		dynamicClient: dynamicClient,
+		retryOptions:  retryOptions,
+	}
+}
+
+// InstallCRDs on a k8s cluster.
+// Returns Output containing results of installation.
+func (i *PreInstaller) InstallCRDs() (Output, error) {
```
Tomasz-Smelcerz-SAP, line 32: Do we consider "UninstallCRDs" operation? I think uninstall should be supported by the library.

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: Will do!

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: So far only logic of resources installation (such as crds or namespaces) has been considered in scope of this task, as this is what was previously done by installer job. I haven't seen anything connected with deinstallation. Tbh your suggestion is a matter of broader Kyma "uninstallation" topic. Even though it was not mentioned in the acceptance criteria I will verify it with PO.

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
...
+func (i *PreInstaller) apply(resourceType resourceType) (o Output, err error) {
+	installationResourcePath := i.cfg.InstallationResourcePath
+	path := fmt.Sprintf("%s/%s", installationResourcePath, resourceType.name)
+
+	rawComponentsDir, err := ioutil.ReadDir(path)
+	if err != nil {
+		return o, err
+	}
+
+	components := findOnlyDirectoriesAmong(rawComponentsDir)
```
Tomasz-Smelcerz-SAP, line 69: What for is a "component" here? I am not sure if we need any "components" in preinstall phase.

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
...
+	if components == nil || len(components) == 0 {
+		i.cfg.Log.Warn("There were no components detected for installation. Skipping.")
+		return o, nil
+	}
+
+	for _, component := range components {
+		componentName := component.Name()
+		i.cfg.Log.Infof("Processing component: %s", componentName)
```
Tomasz-Smelcerz-SAP, line 77: I think that context is missing here: we don't know which resourceType is being processed. So, if we have multiple resource types, we'll have few identical log entries: "Processing component: Foo", one for every resource type.

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: In current approach you can find preinstaller resources in Kyma repository. They are as it follows:
```
- installation
    - resources
        - crds
            - component1
                - file 1
                - file 2
                - ...
                - file n
            - component1
                - file 1
                - file 2
                - ...
                - file n
        - namespaces
            // the same composition as for crds
```
I have introduced a mechanism for creating resources of a specific type. They should always be under appropriate subdirectory placed in `installation/resources`. However, inside these directories resources should be divided by component (for better logical division, ease of component-flagging mechanism that could be introduced etc).
So, to sum up - in this context `component` means either internal Kyma component or external Kyma dependency (which should be still so-called components).

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
...
+
+	output, err := i.apply(resource)
+	if err != nil {
+		return Output{}, err
+	}
+
+	return output, nil
+}
+
+func (i *PreInstaller) apply(resourceType resourceType) (o Output, err error) {
```
Tomasz-Smelcerz-SAP, line 60: I think this function does a bit too much. It is responsible for **finding** resources and for **applying** them.
How about splitting this functionality?
I'd suggest to have two functions:
findResources that does the finding/directory reads and returns a list of objects with fields like {resourceType, component, fileName, path} (only necessary data)
applyResources that is processing the list produced by findResources

I would also split this nested for loop into two private functions each having a single for loop. But I will not insist on that, if the code is simple, a one-level nesting is still OK.

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,127 @@
...
+				o.notInstalled = append(o.notInstalled, file)
+				return o, err // TODO: fail-fast or continue?
+			}
+
+			applied, err := i.applier.Apply(string(resourceData))
+			if err != nil {
+				o.notInstalled = append(o.notInstalled, file)
+				i.cfg.Log.Warnf("Error occurred when processing file %s : %s", resourceName, err.Error())
+			} else {
+				if applied {
```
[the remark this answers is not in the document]
mjakobczyk: In fact it should be considered for adding it to uninstalled ones. Will fix it!

parallel-install/pkg/preinstaller/resourceapplier.go
```diff
@@ -0,0 +1,124 @@
...
+// NewGenericResourceApplier returns a new instance of GenericResourceApplier.
+func NewGenericResourceApplier(log logger.Interface, resourceManager ResourceManager) *GenericResourceApplier {
+	return &GenericResourceApplier{
+		log:             log,
+		decoder:         initializeDecoder(),
+		resourceManager: resourceManager,
+	}
+}
+
+func (c *GenericResourceApplier) Apply(manifest string) (bool, error) {
```
Tomasz-Smelcerz-SAP, line 43: I am missing a clear contract for this method.
It returns "false, nil" in four cases. Three of these are related to parsing/decoding input files. The fourth one is when the resource already exists... This is very dangerous for upgrades. We may silently proceed with an upgrade even if one CRD was not updated to the latest version because it already exists in the cluster.
I would say we should either:
- decide it's an error when resource already exist
- return a specific, unique value that allows the caller to recognize the "resource already exists" condition and leave the decision "what to do next" to the caller.

mjakobczyk: Agree. Contract has been changed. Now it will return error only, so fail-fast approach will be taken.

parallel-install/pkg/preinstaller/preinstaller.go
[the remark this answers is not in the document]
mjakobczyk: Contract has been changed, so this refactor is no longer applicable.

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -0,0 +1,172 @@
...
+
+type resourceInfoInput struct {
+	dirSuffix                string
+	resourceType             string
+	installationResourcePath string
+}
+
+type resourceInfoResult struct {
+	component    string
+	name         string
```
Tomasz-Smelcerz-SAP, line 41: I understand that this holds the file name? I'd rename it to `fileName` then. Just the "name" requires looking at the code to realize "name of what" is stored there.

parallel-install/pkg/preinstaller/resourcemanager_test.go
```diff
@@ -0,0 +1,101 @@
...
+	retryOptions := getTestingRetryOptions(cfg)
+
+	t.Run("should update resource", func(t *testing.T) {
+		// given
+		manager := NewDefaultResourceManager(dynamicClient, retryOptions)
+		resource := unstructured.Unstructured{}
+		resourceSchema := schema.GroupVersionResource{}
+
+		// when
+		err := manager.UpdateRefreshableResource(&resource, resourceSchema)
```
Tomasz-Smelcerz-SAP, line 96: What exactly happens here? The resource is empty and has an empty schema... I am not an expert on dynamic client but I would expect some test data here like `namespace` for example.

parallel-install/pkg/preinstaller/resourcemanager_test.go
```diff
@@ -0,0 +1,101 @@
...
+		}}
+		resourceSchema := schema.GroupVersionResource{}
+		err := manager.CreateResource(&testObj, resourceSchema)
+		assert.NoError(t, err)
+
+		// when
+		obj, err := manager.GetResource(resourceName, resourceSchema)
+
+		// then
+		expectedError := "not found"
```
Tomasz-Smelcerz-SAP, line 74: The test case name suggest we should expect success, not an error.

parallel-install/pkg/preinstaller/resourcemanager_test.go
```diff
@@ -0,0 +1,101 @@
...
+	retryOptions := getTestingRetryOptions(cfg)
+
+	t.Run("should create resource", func(t *testing.T) {
+		// given
+		manager := NewDefaultResourceManager(dynamicClient, retryOptions)
+		resource := unstructured.Unstructured{}
+		resourceSchema := schema.GroupVersionResource{}
+
+		// when
+		err := manager.CreateResource(&resource, resourceSchema)
```
Tomasz-Smelcerz-SAP, line 28: What exactly happens here? The resource is empty and has an empty schema... I am not an expert on dynamic client but I would expect some test data here like a `namespace` for example.

parallel-install/pkg/preinstaller/preinstaller_test.go
```diff
@@ -0,0 +1,230 @@
...
+			dirSuffix:                "crds",
+			installationResourcePath: resourcePath,
+		}
+
+		// when
+		output, err := i.install(input)
+
+		// then
+		assert.NoError(t, err)
+		assert.True(t, len(output.installed) == 2)
```
Tomasz-Smelcerz-SAP, line 41: From my understaning we're not aborting installation on first error, but we continue with remaining resources.
It would be nice to see a test case, where we have: "installed: 3, unInstalled: 1". I don't see such test case?

mjakobczyk: Great point! Added.

parallel-install/pkg/preinstaller/resourcemanager_test.go
mjakobczyk: I have introduced an exemplary resource.

parallel-install/pkg/preinstaller/resourcemanager_test.go
mjakobczyk: I have introduced an exemplary resource.

parallel-install/pkg/preinstaller/preinstaller.go
clebs: The parallel installation API does offer kyma deletion: https://github.com/kyma-incubator/hydroform/blob/6d2f2ab07d912e5a880c858a502ef17e8fddc21d/parallel-install/pkg/deployment/deletion.go#L26

Therefore I guess removing the CRDs needs to be supported.

parallel-install/pkg/preinstaller/resourcemanager.go
```diff
@@ -0,0 +1,116 @@
...
+	return nil
+}
+
+func (c *DefaultResourceManager) GetResource(resourceName string, resourceSchema schema.GroupVersionResource) (obj *unstructured.Unstructured, err error) {
+	obj, err = c.getResource(resourceName, resourceSchema)
+	if err != nil {
+		notFoundError := "not found"
+		matched, _ := regexp.MatchString(notFoundError, err.Error())
+		if matched {
+			c.log.Infof("Resource %s was not found. Skipping additional retries of getting resource.", resourceName)
```
Tomasz-Smelcerz-SAP, line 65:
```suggestion
			c.log.Infof("Resource %s was not found.", resourceName)
```

Tomasz-Smelcerz-SAP (approved):

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: @clebs I have created a follow-up issue for this case - https://github.com/kyma-incubator/hydroform/issues/214

klaudiagrz (approved):

parallel-install/pkg/preinstaller/preinstaller.go
clebs: I am not sure this is fine @mjakobczyk. This PR just broke the current contract of the parallel installation API.
This time it is more or less "OK" t do this because this library is still experimental.

Nevertheless, this will be running everywhere very soon and making breaking changes like this will not be an option.
Please always consider what the current API offers and how changes affect consumers.

Issue #170, opened 2021-01-14
parallel-installation lib error reporting mechanism


strekm:
**Description**

Implement mechanism giving possibility to report errors. This should be consistent with the progress reporting mechanism.

AC:
- client is able to retrieve deploy/uninstallation errors

**Reasons**

Having a consistent way for retrieving errors.

PR #212, opened 2021-02-25
Send error updates

tobiscr:
**Description**

Send errors of worker threads as process updates to caller.

**Related issue(s)**
#170

### DIFF
--- a/parallel-install/pkg/deployment/overrideinterceptors.go
+++ b/parallel-install/pkg/deployment/overrideinterceptors.go
@@ -26,21 +26,6 @@ type OverrideInterceptor interface {
 	Undefined(overrides map[string]interface{}, key string) error
 }
 
-type defaultOverrideInterceptor struct {
-}
-
-func (i *defaultOverrideInterceptor) String(value interface{}, key string) string {
-	return fmt.Sprintf("%v", value)
-}
-
-func (i *defaultOverrideInterceptor) Intercept(value interface{}, key string) (interface{}, error) {
-	return value, nil
-}
-
-func (i *defaultOverrideInterceptor) Undefined(overrides map[string]interface{}, key string) error {
-	return nil
-}
-
 //DomainNameOverrideInterceptor resolves the domain name for the cluster
 type DomainNameOverrideInterceptor struct {
 }
--- a/parallel-install/pkg/deployment/overrideinterceptors.go
+++ b/parallel-install/pkg/deployment/overrideinterceptors.go
@@ -4,7 +4,6 @@ import (
 	"crypto/tls"
 	"encoding/base64"
 	"fmt"
-	"reflect"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -26,21 +25,6 @@ type OverrideInterceptor interface {
 	Undefined(overrides map[string]interface{}, key string) error
 }
 
-type defaultOverrideInterceptor struct {
```
tobiscr, line 29: was dead code

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -150,10 +134,12 @@ func (i *FallbackOverrideInterceptor) Undefined(overrides map[string]interface{}
 		if _, ok := lastProcessedEntry[subKey]; !ok {
 			//sub-element does not exist - add map
 			lastProcessedEntry[subKey] = make(map[string]interface{})
-		} else if reflect.ValueOf(lastProcessedEntry[subKey]).Kind() != reflect.Map {
+		}
+		if _, ok := lastProcessedEntry[subKey].(map[string]interface{}); !ok {
```
tobiscr, line 138: using proper type validation.. thanks @clebs  for the hint ;)

parallel-install/pkg/engine/engine.go
```diff
@@ -64,59 +71,57 @@ type Installation interface {
...
-		return nil, err
-	}
-
+func (e *Engine) Deploy(ctx context.Context) (<-chan components.KymaComponent, <-chan error, error) {
 	//TODO: Size dependent on number of components?
 	statusChan := make(chan components.KymaComponent, 30)
+	errorChan := make(chan error)
 
-	//TODO: Can we avoid this goroutine? Maybe refactor run() so it's non-blocking ?
+	//TODO: Can we avoid this go-routine? Maybe refactor run() so it's non-blocking ?
```
clebs, line 82: There is no misspelling here, golang's threads are called `goroutines`.
Reference: https://golang.org/doc/effective_go#goroutines

parallel-install/pkg/engine/engine.go
```diff
@@ -64,59 +71,57 @@ type Installation interface {
...
 }
 
-func (e *Engine) Deploy(ctx context.Context) (<-chan components.KymaComponent, error) {
-
-	cmps, err := e.componentsProvider.GetComponents()
-	if err != nil {
-		return nil, err
-	}
-
+func (e *Engine) Deploy(ctx context.Context) (<-chan components.KymaComponent, <-chan error, error) {
```
clebs, line 77: Do we really need 2 different channels to send updates to the client? Or would it be possible to use the status channel error field as a replacement for the error channel

parallel-install/pkg/engine/engine.go
tobiscr: Code spell complained and suggest go-routines :D ... will revert it

parallel-install/pkg/engine/engine.go
tobiscr: Unfortunately, we have some rare cases where an error can occur which isn't related to a component, e.g. https://github.com/tobiscr/hydroform/blob/send-error-updates/parallel-install/pkg/engine/engine.go#L121  or https://github.com/tobiscr/hydroform/blob/send-error-updates/parallel-install/pkg/engine/engine.go#L89

parallel-install/pkg/engine/engine.go
tobiscr: FIXED

parallel-install/pkg/engine/engine.go
clebs: I still think it can be avoided, for each link you provided:
1. Since we are returning when the error happens and there is no concurrency at that point inside `run`, `run` can just return an error without a channel. Then `Deploy` just gets it as a normal return value.
2. Same is valid for `Deploy`. The only moment where an error is sent to that channel is before calling `run`. That part of the code can be moved outside of the goroutine (since it anyway happens before `run` it does not benefit from the concurrency. And then we will use the current return error that is not used at all in `Deploy` as well.

tobiscr: Makes sense.. FIXED :)

clebs (approved):

Issue #187, opened 2021-02-08
New installation: Provide Gardener domain upfront


colunira:
**Description**

When running on Gardener, fetch the cluster domain from the `shoot-info` ConfigMap in the `kube-system` namespace and provide it to the deployment as an override `global.domainName`.

**Reasons**

We cannot create overrides for components in components, so we need to get the information about the domain at the beginning of the installation/upgrade.

**Attachments**
https://github.com/kyma-project/kyma/issues/10297

PR #218, opened 2021-03-09
adding gardener domain interceptor

strekm:
**Description**

- adding gardener domain interceptor

**Related issue(s)**
https://github.com/kyma-incubator/hydroform/issues/187

### DIFF
--- a/parallel-install/pkg/deployment/core.go
+++ b/parallel-install/pkg/deployment/core.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/avast/retry-go"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/metadata"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -51,7 +53,13 @@ func newCore(cfg *config.Config, ob *OverridesBuilder, kubeClient kubernetes.Int
 		removeFromComponentList(clList, incompatibleLocalComponents)
 	}
 
-	registerOverridesInterceptors(ob)
+	commonRetryOpts := []retry.Option{
+		retry.Delay(time.Duration(cfg.BackoffInitialIntervalSeconds) * time.Second),
+		retry.Attempts(uint(cfg.BackoffMaxElapsedTimeSeconds / cfg.BackoffInitialIntervalSeconds)),
+		retry.DelayType(retry.FixedDelay),
+	}
+
+	registerOverridesInterceptors(kubeClient, ob, commonRetryOpts, cfg.Log)
 
 	overrides, err := ob.Build()
 	if err != nil {
@@ -157,8 +165,8 @@ func removeFromComponentList(cl *components.ComponentList, componentNames []stri
 	}
 }
 
-func registerOverridesInterceptors(o *OverridesBuilder) {
+func registerOverridesInterceptors(kubeClient kubernetes.Interface, o *OverridesBuilder, retryOptions []retry.Option, log logger.Interface) {
 	//hide certificate data
-	o.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, &DomainNameOverrideInterceptor{})
+	o.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, NewDomainNameOverrideInterceptor(kubeClient, retryOptions, log))
 	o.AddInterceptor([]string{"global.tlsCrt", "global.tlsKey"}, NewCertificateOverrideInterceptor("global.tlsCrt", "global.tlsKey"))
 }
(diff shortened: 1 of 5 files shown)


parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -38,7 +55,48 @@ func (i *DomainNameOverrideInterceptor) Intercept(value interface{}, key string)
...
-	return NewFallbackOverrideInterceptor(localKymaDevDomain).Undefined(overrides, key)
+	domain, err := i.discoverDomain()
+	if err != nil {
+		return err
+	}
+
+	return NewFallbackOverrideInterceptor(domain).Undefined(overrides, key)
+}
+
+func (i *DomainNameOverrideInterceptor) discoverDomain() (string, error) {
```
clebs, line 66: This should probably also run even when the domain is not undefined right? Gardener domain needs to even override user provided custom domains right?

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -25,20 +31,82 @@ type OverrideInterceptor interface {
...
+	}
+
+	if domainName != "" {
+		return domainName, nil
+	}
+	return localKymaDevDomain, nil
+}
+
+func (i *DomainNameOverrideInterceptor) getGardenerDomain() (string, error) {
+	domainName := ""
```
mjakobczyk, line 89: You can use named return values instead of creating a new var.

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -25,20 +31,82 @@ type OverrideInterceptor interface {
...
+		configMap, err := i.kubeClient.CoreV1().ConfigMaps("kube-system").Get(context.TODO(), "shoot-info", metav1.GetOptions{})
+
+		if err != nil {
+			if apierr.IsNotFound(err) {
+				return nil
+			}
+			return err
+		}
+
+		domainName = configMap.Data["domain"]
```
mjakobczyk, line 100: Is there a possibility of `nil` if this property does not exist?

parallel-install/pkg/deployment/core.go
```diff
@@ -51,7 +53,13 @@ func newCore(cfg *config.Config, ob *OverridesBuilder, kubeClient kubernetes.Int
 		removeFromComponentList(clList, incompatibleLocalComponents)
 	}
 
-	registerOverridesInterceptors(ob)
+	commonRetryOpts := []retry.Option{
+		retry.Delay(time.Duration(cfg.BackoffInitialIntervalSeconds) * time.Second),
+		retry.Attempts(uint(cfg.BackoffMaxElapsedTimeSeconds / cfg.BackoffInitialIntervalSeconds)),
```
Tomasz-Smelcerz-SAP, line 58: I have mixed feelings about this... These config params supports `https://github.com/cenkalti/backoff` library, which has a more "declarative" approach to retries. The default values are 30 and 1, meaning your code performs 30 retries - a bit too much IMO. In addition, `BackoffInitialIntervalSeconds` can be zero, which would lead to panic here. I would consider reasonable constant values here, like 3 retries with 2 seconds delay. Another approach is to use `BackoffMaxElapsedTimeSeconds` and divide it by a constant number of retries (like 3 or 5) to get the fixed delay value.

parallel-install/pkg/deployment/core.go
```diff
@@ -157,8 +165,8 @@ func removeFromComponentList(cl *components.ComponentList, componentNames []stri
 	}
 }
 
-func registerOverridesInterceptors(o *OverridesBuilder) {
+func registerOverridesInterceptors(kubeClient kubernetes.Interface, o *OverridesBuilder, retryOptions []retry.Option, log logger.Interface) {
 	//hide certificate data
-	o.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, &DomainNameOverrideInterceptor{})
+	o.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, NewDomainNameOverrideInterceptor(kubeClient, retryOptions, log))
```
Tomasz-Smelcerz-SAP, line 170: From what I can see in the code, the logic (including retries etc.) is performed twice, because we are providing two keys? It would be nice to perform the k8s related operations only once.

parallel-install/pkg/deployment/overrideinterceptors.go
strekm: i don't think so but we can double check and throw an error

parallel-install/pkg/deployment/core.go
strekm: idk it's worth investing in that, cause soon we will get rid of ingress.domainName

Tomasz-Smelcerz-SAP (approved):

parallel-install/pkg/deployment/overrideinterceptors.go
clebs: There is no possibility of `nil`, only empty string ("") since the zero value of a string is "".

Alternatively you can also check like this (more safe):
```suggestion
		domainName, exists := configMap.Data["domain"]
		if !exists {
		  // return an error
		}
```

clebs (approved):

PR #220, opened 2021-03-10
Linters

clebs:
**Description**

- Add golang linters for all hydroform mods
- Currently linters are not mandatory but will be mandatory soon to merge a PR.


tobiscr (approved):

hack/README.md
```diff
@@ -0,0 +1,7 @@
+# Hack directory
+
+This package contains various scripts that are used by Hydroform developers.
+
+## Purpose
```
alexandra-simeonova, line 5:
```suggestion

## Overview

This package contains various scripts that are used by Hydroform developers.

### Purpose
```

function/go.sum
```diff
@@ -1,6 +1,5 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
-cloud.google.com/go v0.38.0 h1:ROfEUZz+Gh5pa62DJWXSaonyu3StP6EA6lPEXPI6mCo=
```
pPrecel, line 3: Have you any idea why was file `go.sum` modify? I can't reproduce it locally

clebs: I just ran `go mod tidy` and it made those changes.
`go.sum` is a generated file, so never edited manually.

suleymanakbas91 (approved):

hack/verify-lint.sh
```diff
@@ -0,0 +1,65 @@
...
+  shout "Run golangci-lint checks"
+  LINTS=(
+    # default golangci-lint lints
+    deadcode errcheck gosimple govet ineffassign staticcheck \
+    structcheck typecheck unused varcheck \
+    # additional lints
+    golint gofmt misspell gochecknoinits unparam scopelint gosec
+  )
+
+  ENABLE=$(sed 's/ /,/g' <<< "${LINTS[@]}")
```
dbadura, line 40: Why the variable is named `ENABLE`.
I would named it as `LINTS`.

clebs: Because it contains the value we pass to the `--enable` flag.
`LINTS` would also be ok of course.

dbadura: I think `LINTS` name is better, because `ENABLE` suggests: `I am a bool`, but in reality I am not :(

clebs: Thing is, 2 lines above we already have the variable named `LINTS` with all the linters in a readable format.
`ENABLE` is just the `LINTS` variable with a format that the `--enable` flag will accept.

I will try to just get rid of the `ENABLE` variable and do the parsing inline.

clebs: Got rid of the `ENABLE` var, so issue solved ;)

clebs: Hi @dbadura, if you do not have other comments, could you please approve the PR?
Thank you.

hack/verify-lint.sh
```diff
@@ -0,0 +1,63 @@
...
+
+source "${CURRENT_DIR}/utilities.sh" || { echo 'Cannot load CI utilities.'; exit 1; }
+
+cleanup() {
+    rm -rf "${TMP_DIR}" || true
+}
+
+trap cleanup EXIT SIGINT
+
+golangci::install() {
```
dbadura, line 21: Do we need to install the tool?
In my opinion it should be in buildpack. 
In case of developer machine, we should add this tool to requirements and add description how to install it or put it on other script

hack/verify-lint.sh
```diff
@@ -0,0 +1,63 @@
...
+    # default golangci-lint lints
+    deadcode errcheck gosimple govet ineffassign staticcheck \
+    structcheck typecheck unused varcheck \
+    # additional lints
+    golint gofmt misspell gochecknoinits unparam scopelint gosec
+  )
+
+  echo "Checks: ${LINTS[*]}"
+  cd $ROOT_PATH
+  golangci-lint --disable-all --enable="$(sed 's/ /,/g' <<< "${LINTS[@]}")" --timeout=10m run --config $CURRENT_DIR/.golangci.yml
```
dbadura, line 42: Can we remove this ugly sed and just use `,` in `LINTS`?

hack/utilities.sh
```diff
@@ -0,0 +1,20 @@
+#
+# Library of useful utilities for CI purposes.
+#
+
+readonly RED='\033[0;31m'
+readonly GREEN='\033[0;32m'
+readonly INVERTED='\033[7m'
+readonly NC='\033[0m' # No Color
```
dbadura, line 8: Do we need to use colors in terminal?

hack/verify-lint.sh
dbadura: Generally we can get rid of this script and put the execution of lint in makefile.

hack/verify-lint.sh
clebs: This script is built so that it works on both CI and developer machines.
But there is also a flag that can be passed to skip the installation of `golangci lint`, so if the environment already has it, the flag can be passed and it will not be installed.

hack/verify-lint.sh
clebs: The idea here was that `LINTS` is easy to read by a human so that we can:
- Easily see/add/remove which linters we are running.
- Document why we have specific linters.

I thought of having this in the `Makefile` but found it useful to be able to have this working out of the box without previous setup and more detailed output.

hack/utilities.sh
clebs: This was done imitating the output format from all other scripts we have in kyma and on prow. Helps having a nicer output and colors help driving the attention to the important stuff.

hack/verify-lint.sh
dbadura: By default I would turn off installing.

dbadura: What do u think to turn it off by default?
Because other tools are also not installed every time.

clebs: In hydroform we are calling the `before-commit.sh` script with the `ci` flag. I will change it so that when using that flag `golangci-lint` is not installed per default.

Disper:
Hi,
we had a look at this and also discussed this with the Wookiee team, sharing the feedback from the meeting we're having.  

We see the value in having linters - thank you for raising the topic! 

We would like to spend more time on this - e.g. review the rules and see which of them could be added as global rules and which could be more restrictive on our team level. We also raised some comments in the PR - in case it would be merged, it would be great that we could revisit those points later as well. We're now focusing on Kyma 2.0 goal, that's the reason that we're asking for the possibility to revisit those later or maybe even propose something :).

hack/verify-lint.sh
clebs: Added linter to the `golang-toolbox` image: https://github.com/kyma-project/test-infra/pull/3441#event-4495789770
Updated prow jobs using linter: https://github.com/kyma-project/test-infra/pull/3447

Disabled installation of `golangci-lint` on CI flagged executions. You can see on the pipeline that the linter is not being downloaded anymore.

clebs:
Hi @Disper,

of course, this first PR is the pilot and linters are only optional. Next phase would be to go through the issues, fix them and identify rules that might need to be removed or added.

The goal is to start doing this and define this as a standard in our development process across Kyma, otherwise it will never happen. But of course, this will need polishing over time.

Regarding the comments, all of them have been addressed as far as I know. Nevertheless, this is a living topic and anyone can propose/make changes anytime.

hack/verify-lint.sh
dbadura: Several reason why the script shouldn't download the tool by default:
- it makes this script slower for developers. 
- In case that the download page will be unavailable and we have this tool locally, the developer will stuck because the linter target want to install something.
- It's violation of single responsibility principle, because this downloads and validate, but the name is `lint-something`
- this project is written in Go, why we don't always download Go when compiling it?

If you want to download something, you can put it in seperate target and in `linter` target. 
It makes more sense to check if the tool is available. When there is no such tool print error and solution how to resolve it.

dbadura: Installing linter is also one time action.

clebs: We can add the check to see if the tool is installed yes, still using an older version of the linter might have different results locally and on the pipeline. And that will be very confusing.

What we can do is printing out the installation command with instructions in case the binary is not found or the version is not the expected one.

clebs: Hi @dbadura,

I have changed the script so that it now verifies the golangci-lint installation and prints install/update instructions if not correct.
Now the developer can install once himself the linter or update it if we increase the minimum version.

dbadura: Looks much better now ;)

dbadura (approved):

PR #223, opened 2021-03-11
Adapt function CLI to new event mesh

m00g3n:
**Description**

- Adapt function CLI to new event mesh

**Related issue(s)**

See also: https://github.com/kyma-project/cli/issues/770

### DIFF
--- a/function/pkg/operator/subscription_operator.go
+++ b/function/pkg/operator/subscription_operator.go
@@ -1,0 +1,82 @@
+package operator
+
+import (
+	"context"
+
+	"github.com/kyma-incubator/hydroform/function/pkg/client"
+	"github.com/kyma-incubator/hydroform/function/pkg/resources/types"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type subscriptionOperator struct {
+	fnRef functionReference
+	items []unstructured.Unstructured
+	client.Client
+}
+
+func NewSubscriptionOperator(c client.Client, fnName, fnNamespace string, u ...unstructured.Unstructured) Operator {
+	return &subscriptionOperator{
+		Client: c,
+		items:  u,
+		fnRef: functionReference{
+			name:      fnName,
+			namespace: fnNamespace,
+		},
+	}
+}
+
+func (t subscriptionOperator) Apply(ctx context.Context, opts ApplyOptions) error {
+	predicate := buildMatchRemovedSubscriptionsPredicate(t.fnRef, t.items)
+
+	if err := wipeRemoved(ctx, t.Client, predicate, opts.Options); err != nil {
+		return err
+	}
+	// apply all subscriptions
+	for _, u := range t.items {
+		u.SetOwnerReferences(opts.OwnerReferences)
+		// fire pre callbacks
+		if err := fireCallbacks(&u, nil, opts.Pre...); err != nil {
+			return err
+		}
+		new1, statusEntry, err := applyObject(ctx, t.Client, u, opts.DryRun)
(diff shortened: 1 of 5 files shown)


function/pkg/operator/generic_operator.go
```diff
@@ -9,21 +9,22 @@ import (
 )
 
 var (
+	//FIXME rename to GVR...
```
rJankowski93, line 12: I think '//TODO rename to GVR...' is better because IDEs find word 'TODO' and inform dev about it.

function/pkg/operator/subscription_operator.go
```diff
@@ -2,27 +2,22 @@ package operator
...
-type triggersOperator struct {
+type subscriptionOperator struct {
 	fnRef functionReference
 	items []unstructured.Unstructured
 	client.Client
 }
 
-func NewTriggersOperator(c client.Client, fnName, fnNamespace string, u ...unstructured.Unstructured) Operator {
-	return &triggersOperator{
+func NewSubscriptionsOperator(c client.Client, fnName, fnNamespace string, u ...unstructured.Unstructured) Operator {
```
rJankowski93, line 19: Sometimes you use 'subscriptions' , sometimes you use 'subscription'. 
I think we should use one form.

function/pkg/resources/unstructured/subscription.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/kyma-incubator/hydroform/function/pkg/resources/types"
+	"github.com/kyma-incubator/hydroform/function/pkg/workspace"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const (
+	apiVersionSubscription = "eventing.kyma-project.io/v1alpha1"
+)
```
rJankowski93, line 15:
```suggestion
const apiVersionSubscription = "eventing.kyma-project.io/v1alpha1"
```

According with function.go and git_repository.go

function/pkg/resources/unstructured/subscription.go
```diff
@@ -0,0 +1,126 @@
...
+
+type toUnstructured func(obj interface{}) (map[string]interface{}, error)
+
+func NewSubscriptions(cfg workspace.Cfg) ([]unstructured.Unstructured, error) {
+	return newSubscriptions(cfg, runtime.DefaultUnstructuredConverter.ToUnstructured)
+}
+
+func newSubscriptions(cfg workspace.Cfg, f toUnstructured) ([]unstructured.Unstructured, error) {
+	var list []unstructured.Unstructured
+	//TODO remove http protocol once it will be fixed in eventng
```
rJankowski93, line 45:
```suggestion
	//TODO remove http protocol once it will be fixed in eventing
```

function/pkg/resources/unstructured/subscription.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if subscriptionName == "" {
+		elems := append([]string{functionName}, subscriptionSources...)
+		subscriptionName = joinNonEmpty(elems, "-")
+	}
+	return subscriptionName
+}
+
+func filterSources(s workspace.Subscription) []string {
+	result := []string{}
```
rJankowski93, line 99:
```suggestion
	var result []string
```
String can't be nil

function/pkg/operator/generic_operator.go
m00g3n: I think it should be also visible with `//FIXME` (at least for vscode)
image

rJankowski93 (approved):

tgorgol (approved):

PR #237, opened 2021-04-01
Callback functions in parallel-install library

akgalwas:
**Description**

This PR contains proposal for API change.

**Motivation**
The current Deployer API has blocking nature and publish updates on the channel.

It has the following consequences:
- Caller needs to launch an additional go routine to process status updates.
- Caller needs to allocate channel and close it when `StartKymaDeployment` completed.

It violates the following rules:
- Keep your APIs Concurrency-Free. Concurrency is an implementation detail and as such should be hidden. 
- In most cases channel should be closed by a process that writes data to the channel.

Recently I read a good explanation for [Leave Concurrency to the caller proverb](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_leave_concurrency_to_the_caller) that provoked me to prepare this PR.     

The proposed changes include:
- Providing a callback function instead of a channel.
- Calling the callback function when update state event is triggered.

The proposed changes have the following advantages:
- Simpler usage of the API. User needs to provide a simple function only.
- No need for creating a new go routine and channel. This is particularly important for runtime provisioning use case in the Control Plane. The Provisioner handles a number of installations concurrently and the new API implicates lesser resource consumption.

The solution is also idiomatic as you can find similar approach in Go standard library (e.g. [filepath.Walk]( https://golang.org/pkg/path/filepath/#Walk)). 

**Using the API**

**Current implementation:**
```go
func deploy() {
       log := logger.NewLogger(*verbose)
       
        var progressCh chan deployment.ProcessUpdate
	progressCh = make(chan deployment.ProcessUpdate)
	ctx := renderProgress(progressCh, log)
	defer close(progressCh)

	deployer, err := deployment.NewDeployment(installationCfg, builder, kubeClient, progressCh)
	if err != nil {
		log.Fatalf("Failed to create installer: %v", err)
	}
	err = deployer.StartKymaDeployment()
	if err != nil {
		log.Errorf("Failed to deploy Kyma: %v", err)
	} else {
		log.Info("Kyma deployed!")
	}
}

func renderProgress(progressCh chan deployment.ProcessUpdate, log logger.Interface) {
	
      showCompStatus := func(comp components.KymaComponent) {
		if comp.Name != "" {
			log.Infof("Status of component '%s': %s", comp.Name, comp.Status)
		}
	}
	go func() {
		for update := range progressCh {
			switch update.Event {
			case deployment.ProcessStart:
				log.Infof("Starting installation phase '%s'", update.Phase)
			case deployment.ProcessRunning:
				showCompStatus(update.Component)
			case deployment.ProcessFinished:
				log.Infof("Finished installation phase '%s' successfully", update.Phase)
			default:
				//any failure case
				log.Infof("Process failed in phase '%s' with error state '%s':", update.Phase, update.Event)
				showCompStatus(update.Component)
			}
		}
	}()
}
```

**Proposed implementation:**
```go
func deploy() {
        log := logger.NewLogger(*verbose)
        renderProgress := getRenderProgressFunc(log)

	deployer, err := deployment.NewDeployment(installationCfg, builder, kubeClient, renderProgress)
	if err != nil {
		log.Fatalf("Failed to create installer: %v", err)
	}

	err = deployer.StartKymaDeployment()
	if err != nil {
		log.Errorf("Failed to deploy Kyma: %v", err)
	} else {
		log.Info("Kyma deployed!")
	}
}

func getRenderProgressFunc(log logger.Interface) func (deployment.ProcessUpdate){
        return func (update deployment.ProcessUpdate) {
	    showCompStatus := func(comp components.KymaComponent) {
	     	  if comp.Name != "" {
			log.Infof("Status of component '%s': %s", comp.Name, comp.Status)
		  }
     	    }

	    switch update.Event {
	    case deployment.ProcessStart:
		    log.Infof("Starting installation phase '%s'", update.Phase)
	     case deployment.ProcessRunning:
		  showCompStatus(update.Component)
	    case deployment.ProcessFinished:
		  log.Infof("Finished installation phase '%s' successfully", update.Phase)
	    default:
		  //any failure case
		  log.Infof("Process failed in phase '%s' with error state '%s':", update.Phase, update.Event)
		  showCompStatus(update.Component)
	    }
      }
}
```

 Related issue: #253


jasiu001 (dismissed):

parallel-install/pkg/deployment/deployment_test.go
```diff
@@ -16,44 +17,46 @@ import (
 	"time"
 )
 
+func callbackUpdate(update ProcessUpdate) {
```
clebs, line 20: What do you think about providing a function like this on the `deployment` pkg as a simple default that clients can use and learn from.
Someting like:
```go
    func DefaultUpdater() func(update ProcessUpdate) {
        return func(update ProcessUpdate) {
                 // Do simple logging of updates.
        }
    }
```

It would be used as follows by a client:
```go
installer, err := deployment.NewDeployment(installationCfg, overrides, cmd.K8s.Static(), deployment.DefaultUpdater())
```

clebs (approved):
LGTM, just an optional suggestion that might be useful.

parallel-install/example/example.go
```diff
@@ -1,9 +1,9 @@
 package main
 
 import (
-	"context"
 	"flag"
 	"fmt"
+	log "github.com/sirupsen/logrus"
```
clebs, line 6: Do we really need this dependency or can we just use golang's `log` package?
Seems a bit too much for just a small example and a unit test.

koala7659: That's a good point. I will remove it.

clebs: Just use the log package from the go standard lib:
```go
import "log"
```

No need to change anything on the code.

parallel-install/pkg/deployment/deployment_test.go
koala7659: Good idea. I will add such a function

parallel-install/example/example.go
koala7659: I agree but I noticed  that in this file we already use logger from our internal package `logger` `(pkg\logger\logger.go)`
It is instantiated in line 31.  

`log = logger.NewLogger(*verbose)`

Please check the file now. Now it we have only one logger used in this file.

clebs (approved):

PR #251, opened 2021-04-09
fix helm kubeconfig

strekm:
**Description**

- fix helm kubeconfig

**Related issue(s)**


parallel-install/example/example.go
```diff
@@ -82,6 +82,7 @@ func main() {
 		ComponentList:                 compList,
 		ResourcePath:                  fmt.Sprintf("%s/src/github.com/kyma-project/kyma/resources", goPath),
 		InstallationResourcePath:      fmt.Sprintf("%s/src/github.com/kyma-project/kyma/installation/resources", goPath),
+		KubeconfigPath:                *kubeconfigPath,
```
tobiscr, line 85: We should also the kubec-client in the API interface, otherwise we path the kube-client and the kubeconfig to the installer API:

* https://github.com/kyma-incubator/hydroform/blob/master/parallel-install/pkg/deployment/core.go#L29
* https://github.com/kyma-incubator/hydroform/blob/master/parallel-install/pkg/deployment/deployment.go#L27
* https://github.com/kyma-incubator/hydroform/blob/master/parallel-install/pkg/deployment/deletion.go#L30

tobiscr (changes requested):

tobiscr (dismissed):

tobiscr (approved):

PR #257, opened 2021-04-13
Introduce KubeConfigManager

mjakobczyk:
**Description**

- expose `kubeconfigcontent` property
- add `KubeconfigSource` struct for input data
- add `Path` and `RestConfig` utility functions for managing Kubeconfig

**Related issue(s)**

Implements https://github.com/kyma-project/hydroform/issues/256


parallel-install/example/example.go
```diff
@@ -30,8 +32,9 @@ func main() {
 
 	log := logger.NewLogger(*verbose)
 
-	if kubeconfigPath == nil || *kubeconfigPath == "" {
-		log.Fatal("kubeconfig is required")
+	kubeConfigManager, err := config.NewKubeConfigManager(kubeconfigPath, kubeconfigContent)
```
strekm, line 35: is it necessary to expose this manager to our clients? in my opinion it should be hidden and resolved under the hood despite what clients set: kubeconfigPath / kubeconfigContent

mjakobczyk: It is not. Now `KubeconfigManager` works on both parameters under the hood in every component that needs kubeClient or dynamicClient.

parallel-install/example/example.go
```diff
@@ -30,8 +31,9 @@ func main() {
 
 	log := logger.NewLogger(*verbose)
 
-	if kubeconfigPath == nil || *kubeconfigPath == "" {
-		log.Fatal("kubeconfig is required")
+	if (kubeconfigPath == nil || *kubeconfigPath == "") &&
+		(kubeconfigRaw == nil || *kubeconfigRaw == "") {
+		log.Fatal("either kubeconfig or kubeconfigcontent property has to be set")
```
Tomasz-Smelcerz-SAP, line 36: it's `kubeconfigraw` or `kubeconfigcontent`?

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -25,19 +25,19 @@ package preinstaller
...
 	"os"
 )
 
 // Config defines configuration values for the PreInstaller.
 type Config struct {
 	InstallationResourcePath string           //Path to the installation resources.
 	Log                      logger.Interface //Logger to be used
-	KubeconfigPath           string
+	KubeconfigPath           string           // Path to the kubeconfig file
+	KubeconfigRaw            string           // Content of the kubeconfig file
```
Tomasz-Smelcerz-SAP, line 40: Consider renaming to `KubeconfigContent` to be consistent with naming in `Config` type

parallel-install/pkg/preinstaller/resourcemanager.go
```diff
@@ -38,8 +37,13 @@ type DefaultResourceManager struct {
 }
 
 // NewResourceManager creates a new instance of ResourceManager.
-func NewDefaultResourceManager(kubeconfigPath string, log logger.Interface, retryOptions []retry.Option) (*DefaultResourceManager, error) {
-	restConfig, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+func NewDefaultResourceManager(kubeconfigPath, kubeconfigRaw string, log logger.Interface, retryOptions []retry.Option) (*DefaultResourceManager, error) {
```
Tomasz-Smelcerz-SAP, line 40: Why don't we just pass directly`*rest.Config` as a parameter?

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -25,19 +25,19 @@ package preinstaller
...
-
 	"os"
 )
 
 // Config defines configuration values for the PreInstaller.
 type Config struct {
 	InstallationResourcePath string           //Path to the installation resources.
 	Log                      logger.Interface //Logger to be used
-	KubeconfigPath           string
+	KubeconfigPath           string           // Path to the kubeconfig file
```
Tomasz-Smelcerz-SAP, line 39: perhaps it's easier to inject `*rest.Config` here instead?

parallel-install/pkg/config/kubeconfigmanager.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Path returns a path to the kubeconfig file.
+func (k *kubeConfigManager) Path() (string, error) {
+	return k.resolvePath()
+}
+
+// Config returns a kubeconfig REST Config used by k8s clients.
+func (k *kubeConfigManager) Config() (*rest.Config, error) {
+	if exists(&k.path) {
+		resolvedPath, err := k.resolvePath()
```
strekm, line 49: i think that in this case we should use what we have instead of creating file if path is empty. if client will provide kubeconfig content then there's no reason to create file for him. we only need file in case of helm scenario

parallel-install/example/example.go
strekm: it should be kubeconfigContent i would say

parallel-install/pkg/config/kubeconfigmanager.go
strekm: and i just noticed that creation of file won't be executed in this case

parallel-install/pkg/config/kubeconfigmanager.go
```diff
@@ -0,0 +1,89 @@
...
+			return nil, err
+		}
+
+		return clientcmd.BuildConfigFromFlags("", resolvedPath)
+	} else {
+		return clientcmd.RESTConfigFromKubeConfig([]byte(k.content))
+	}
+}
+
+func (k *kubeConfigManager) resolvePath() (path string, err error) {
```
strekm, line 60: this is a bit confusing for me cause i would say that creating temp file shouldn't be responsibility of this method. especially in case it's usage in Config() method

parallel-install/pkg/config/kubeconfigmanager.go
strekm: shouldn't we use just k.path in this case?

parallel-install/pkg/preinstaller/resourcemanager.go
mjakobczyk: Currently all objects handle `rest.Config` on their own and create it internally.

parallel-install/example/example.go
mjakobczyk: At first I called It `kubeconfigcontent` but later I have spotted Magda mentioning `kubeconfigraw` as a potential candidate to being exposed as a property. I have simply not changed the error message.

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: For now the trend is to pass all kubeconfigs. I think that library's consumers want to simply pass config properties, such as kubeconfigs.

parallel-install/pkg/config/kubeconfigmanager.go
mjakobczyk: To simplify things it can be changed to `k.path`.

parallel-install/pkg/config/kubeconfigmanager.go
mjakobczyk: Ok - a separate function for creating a file and the second one to create a final path seems decent.

parallel-install/pkg/preinstaller/preinstaller.go
strekm: yeah ofc we can do that buuuuuut helm need kubeconfigpath and that's the problem

jasiu001:
LGTM

jasiu001 (dismissed):

Tomasz-Smelcerz-SAP (approved):

strekm (dismissed):
/meow

parallel-install/example/example.go
```diff
@@ -30,8 +29,9 @@ func main() {
 
 	log := logger.NewLogger(*verbose)
 
-	if kubeconfigPath == nil || *kubeconfigPath == "" {
-		log.Fatal("kubeconfig is required")
+	if (kubeconfigPath == nil || *kubeconfigPath == "") &&
+		(kubeconfigContent == nil || *kubeconfigContent == "") {
```
clebs, line 33: I don't think you need to check for `nil` here. The `flag.String()` function will always return a non `nil` string pointer (either passed flag or pointer to empty string). In this case if the flag is not passed, an empty string, but the pointer will never be nil.

parallel-install/pkg/config/config_test.go
```diff
@@ -30,24 +30,16 @@ func Test_ValidateDeletion(t *testing.T) {
 	t.Run("Happy path", func(t *testing.T) {
 		fpath := filePath(t)
 		config = Config{
-			WorkersCount:   1,
-			ComponentList:  newComponentList(t),
-			KubeconfigPath: filepath.Dir(fpath),
+			WorkersCount:  1,
+			ComponentList: newComponentList(t),
+			KubeconfigSource: KubeconfigSource{
+				Path:    filepath.Dir(fpath),
```
clebs, line 36: Is there a reason to strip the filename from the path here?

parallel-install/pkg/config/kubeconfigmanager.go
```diff
@@ -0,0 +1,107 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+const (
+	temporaryFilePattern = "kubeconfig-*.yaml"
+)
+
+// kubeConfigManager manages resolving kubeconfig from two possible sources: a file path or a kubeconfig content.
+type kubeConfigManager struct {
```
clebs, line 17: I think this type is redundant if we already have `KubeconfigSource`.

All logic here can be attached to `KubeconfigSource` without having unnecessary layers and abstractions. Keep it simple :)

parallel-install/pkg/config/kubeconfigmanager.go
```diff
@@ -0,0 +1,107 @@
...
+	kubeconfigSource KubeconfigSource
+	temporaryPath    string // Mutable!
```
clebs, line 19: This field can be deleted and replaced with a variable in the only function it is used.
This would also automatically make it thread safe, which is not the case at the moment.

parallel-install/example/example.go
mjakobczyk: Actually it is just in `example.go` and the `version` is checked in the same manner. Is it a problem to leave it this way here?

parallel-install/pkg/config/config_test.go
mjakobczyk: It used to do so before the refactor when only path was provided, so I left it the same but now using `KubeconfigSource`.

parallel-install/example/example.go
clebs: Not a problem, just less code => easier to read.

parallel-install/pkg/config/kubeconfigmanager.go
mjakobczyk: According to what Magda said https://github.com/kyma-project/hydroform/pull/257#discussion_r612183716 she does not want that struct to be explicitly set in the installation config. As a result, it is being resolved under the hood. On the other hand, I find it pretty much "ugly" to pass two parameters to every method that need kubeconfig. As a result, there is a simple struct that aggregates these two parameters, so we pass it around anywhere.

parallel-install/pkg/config/config_test.go
clebs: What happens if for example the user provides a path like this?
`/some/path/totally-random-filename.yaml`

Will the implementation of `KubeconfigSource` still work?

parallel-install/pkg/config/kubeconfigmanager.go
mjakobczyk: We create a separate `KubeconfigManager` for every client that needs kubeconfig at the moment. For a helm use case - it is even a matter of a method. It means, when you want to deploy a release using our helm client then it will always create a new manager with a single temporary file and remove it after the deploy is done. The same to release uninstall. We do not change the path in any way when it is passed to the method that uses concurrency under the hood.

clebs (changes requested):
Please see single comments on the code.

parallel-install/pkg/config/kubeconfigmanager.go
clebs: Maybe you understood me wrong. What I am proposing, is to keep `KubeconfigPath` and delete `KubeconfigManager`.

Then move all the methods from `KubeconfigManager` to `KubeconfigPath`.

parallel-install/pkg/config/kubeconfigmanager.go
clebs: Still, why have an unnecessary field in the type?

parallel-install/pkg/config/kubeconfigmanager.go
Tomasz-Smelcerz-SAP: Personally I pefer the exact opposite. The external type is simple placeholder. The internal type contains logic, and we are free to change the contract. If we expose this logic as part of public API, then ANY change to this logic would be a breaking change in our public API. We don't want that. We want to be free to change the logic without affecting the users.

strekm (approved):

clebs (approved):

Issue #234, opened 2021-03-31
Patching coredns in the installation library


strekm:
**Description**

To ensure communication we need to patch coredns as follows:
```
data:
  Corefile: |
    .:53 {
        errors
        health
        rewrite name regex (.*)\.local\.kyma\.dev istio-ingressgateway.istio-system.svc.cluster.local
        ready
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
        hosts /etc/coredns/NodeHosts {
          reload 1s
          fallthrough
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```
AC:
- domain shouldn't be hardcoded (we will be needing similar patch for other scenario)
- logic should be implemented in the installation lib interceptor
- if coredns is not installed log it and proceed with installation
- documentation with a note that kyma will fully work only with coredns (new documentation)

**Reasons**

For kyma.local.dev domain we need to patch coredns to ensure traffic.

**Attachments**

PR #270, opened 2021-04-28
Coredns patch

werdes72:
**Description**

- patch coredns configuration with proper domain name during installation

See #234

### DIFF
--- a/parallel-install/example/example.go
+++ b/parallel-install/example/example.go
@@ -14,7 +14,6 @@ import (
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/helm"
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/preinstaller"
-	"k8s.io/client-go/kubernetes"
 )
 
 var log *logger.Logger
@@ -106,27 +105,6 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create Kyma pre-installer: %v", err)
 	}
-
-	// TODO: REMOVE, JUST FOR TESTING!
-	restConfig, err := config.RestConfig(installationCfg.KubeconfigSource)
-	if err != nil {
-		log.Fatal("aaa")
-	}
-	kubeClient, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		log.Fatal(("bbb"))
-	}
-
-	builder.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, deployment.NewDomainNameOverrideInterceptor(kubeClient, log))
-	builder.AddInterceptor([]string{"global.tlsCrt", "global.tlsKey"}, deployment.NewCertificateOverrideInterceptor("global.tlsCrt", "global.tlsKey", kubeClient))
-	overrides, err := builder.Build()
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	log.Info(overrides)
-
-	return
 
 	result, err := preInstaller.InstallCRDs()
 	if err != nil || len(result.NotInstalled) > 0 {
--- a/parallel-install/example/example.go
+++ b/parallel-install/example/example.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/helm"
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/preinstaller"
+	"k8s.io/client-go/kubernetes"
 )
 
 var log *logger.Logger
@@ -105,6 +106,27 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to create Kyma pre-installer: %v", err)
 	}
+
+	// TODO: REMOVE, JUST FOR TESTING!
+	restConfig, err := config.RestConfig(installationCfg.KubeconfigSource)
+	if err != nil {
+		log.Fatal("aaa")
+	}
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		log.Fatal(("bbb"))
+	}
+
+	builder.AddInterceptor([]string{"global.domainName", "global.ingress.domainName"}, deployment.NewDomainNameOverrideInterceptor(kubeClient, log))
+	builder.AddInterceptor([]string{"global.tlsCrt", "global.tlsKey"}, deployment.NewCertificateOverrideInterceptor("global.tlsCrt", "global.tlsKey", kubeClient))
+	overrides, err := builder.Build()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Info(overrides)
+
+	return
 
 	result, err := preInstaller.InstallCRDs()
 	if err != nil || len(result.NotInstalled) > 0 {
--- a/parallel-install/pkg/deployment/coredns.go
+++ b/parallel-install/pkg/deployment/coredns.go
@@ -1,0 +1,124 @@
+package deployment
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"html/template"
+	"strings"
+
+	"github.com/avast/retry-go"
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
+	v1 "k8s.io/api/core/v1"
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+var coreDNSPatchTemplate = `
+.:53 {
+    errors
+    health
+    rewrite name regex {{ .DomainName}} istio-ingressgateway.istio-system.svc.cluster.local
+    ready
+    kubernetes cluster.local in-addr.arpa ip6.arpa {
+      pods insecure
+      fallthrough in-addr.arpa ip6.arpa
+    }
+    hosts /etc/coredns/NodeHosts {
+      reload 1s
+      fallthrough
+    }
+    prometheus :9153
+    forward . /etc/resolv.conf
+    cache 30
+    loop
+    reload
+    loadbalance
+}
+`
+
+// CoreDNSPatch contains values to fill template with
+type CoreDNSPatch struct {
+	DomainName string
+}
+
+// patchCoreDNS takes kubeclient and cluster domain as a parameter to patch coredns config
+// e.g. domainName: `(.*)\.local\.kyma\.dev`
+func patchCoreDNS(kubeClient kubernetes.Interface, domainName string, log logger.Interface) (cm v1.ConfigMap, err error) {
+	// TODO: Refactor
+	err = retry.Do(func() error {
+		_, err := kubeClient.AppsV1().Deployments("kube-system").Get(context.TODO(), "coredns", metav1.GetOptions{})
+		if err != nil {
+			if apierr.IsNotFound(err) {
+				log.Info("CoreDNS not found, skipping CoreDNS config patch")
+				return nil
+			}
+			return err
+		}
+
+		coreFile, err := generateCorefile(domainName)
+		if err != nil {
+			return err
+		}
+		configMaps := kubeClient.CoreV1().ConfigMaps("kube-system")
+		coreDNSConfigMap, err := configMaps.Get(context.TODO(), "coredns", metav1.GetOptions{})
+		if err != nil {
+			if apierr.IsNotFound(err) {
+				log.Info("Corefile not found, creating new CoreDNS config")
+				newCM, err := configMaps.Create(context.TODO(), getNewCoreDNSConfigMap(coreFile), metav1.CreateOptions{})
+				cm = *newCM
+				if err != nil {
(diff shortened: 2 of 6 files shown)


parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -28,6 +33,40 @@ const (
...
 )
 
+var retryOptions = []retry.Option{
+	retry.Delay(2 * time.Second),
+	retry.Attempts(3),
+	retry.DelayType(retry.FixedDelay),
+}
+
+var coreDNSPatchTemplate = `
```
Tomasz-Smelcerz-SAP, line 42: please move coreDNS related things to a separate file in the same package. Then it's nicely separated from other things here and can be moved easily if necessary.

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -100,10 +139,18 @@ func (i *DomainNameOverrideInterceptor) getDomainName() (domainName string, err
 		return "", err
 	}
 	if domainName != "" {
+		err = patchCoreDNS(i.kubeClient, `(.*)\.local\.kyma\.devaaaa`, i.log)
```
Tomasz-Smelcerz-SAP, line 142: We are patching anyway few lines below with a different argument. Perhaps we can simplify and this invocation can be avoided.

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -121,6 +168,75 @@ func (i *DomainNameOverrideInterceptor) findLocalDomain() (domainName string, er
...
+	coreDNSPatch := CoreDNSPatch{DomainName: domainName}
+	patchTemplate := template.Must(template.New("").Parse(coreDNSPatchTemplate))
+	patchBuffer := new(bytes.Buffer)
+	if err = patchTemplate.Execute(patchBuffer, coreDNSPatch); err != nil {
+		return err
+	}
+
+	// TODO: Refactor
+	err = retry.Do(func() error {
+		_, err := kubeClient.AppsV1().Deployments("kube-system").Get(context.TODO(), "coredns", metav1.GetOptions{})
```
Tomasz-Smelcerz-SAP, line 183: Extract to a separate function that is responsible for coreDNS detection.

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -121,6 +168,75 @@ func (i *DomainNameOverrideInterceptor) findLocalDomain() (domainName string, er
...
+		if err != nil {
+			if apierr.IsNotFound(err) {
+				log.Info("CoreDNS not found, skipping CoreDNS config patch")
+				return nil
+			}
+			return err
+		}
+
+		configMaps := kubeClient.CoreV1().ConfigMaps("kube-system")
```
Tomasz-Smelcerz-SAP, line 192: how about refactoring as follows (don't pay attention to naming ATM)
```
exists, err := findCoreDNSMap(...)
if err {
  return err
}
if exists {
   err = patchCoreDNSMap(...)
} else {
  err = createCoreDNSMap(...)
}
return err
```

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Moved to separate file.

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Done

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Done

parallel-install/example/example.go
```diff
@@ -106,6 +107,27 @@ func main() {
 		log.Fatalf("Failed to create Kyma pre-installer: %v", err)
 	}
 
+	// TODO: REMOVE, JUST FOR TESTING!
```
mjakobczyk, line 110: Make sure to delete it before merging.

parallel-install/pkg/deployment/coredns.go
```diff
@@ -0,0 +1,152 @@
...
+    forward . /etc/resolv.conf
+    cache 30
+    loop
+    reload
+    loadbalance
+}
+`
+
+// CoreDNSPatch contains values to fill template with
+type CoreDNSPatch struct {
```
mjakobczyk, line 44: Does this type has to be exported? From what I can see you use it only for internal purpose and do not expose any patching strategy to the library users.

parallel-install/pkg/deployment/coredns.go
```diff
@@ -0,0 +1,152 @@
...
+	DomainName string
+}
+
+// patchCoreDNS takes kubeclient and cluster domain as a parameter to patch coredns config
+// e.g. domainName: `(.*)\.local\.kyma\.dev`
+func patchCoreDNS(kubeClient kubernetes.Interface, domainName string, log logger.Interface) (cm v1.ConfigMap, err error) {
+	// TODO: Refactor
```
mjakobczyk, line 51: Are you done with this TODO?

parallel-install/pkg/deployment/coredns_test.go
```diff
@@ -0,0 +1,64 @@
...
+		// when
+		cm, err := patchCoreDNS(kubeClient, `(.*)\.loacal\.kyma\.dev`, log)
+
+		// then
+		require.NoError(t, err)
+		require.Empty(t, cm.Data)
+	})
+}
+
+func fakeCoreDNSCM() *v1.ConfigMap {
```
mjakobczyk, line 30: I don't like the name while you combine two different words with uppercase - DNS and CM. Could you name it differently to make it more readable, like: `fakeCoreDnsCm` or `fakeCoreDNScm`?

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -28,6 +28,12 @@ const (
...
 )
 
+var retryOptions = []retry.Option{
```
mjakobczyk, line 31: Should we create package-wide variables? Or could it become a constant?

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -100,10 +106,18 @@ func (i *DomainNameOverrideInterceptor) getDomainName() (domainName string, err
 		return "", err
 	}
 	if domainName != "" {
+		_, err = patchCoreDNS(i.kubeClient, `(.*)\.local\.kyma\.dev`, i.log)
```
mjakobczyk, line 109: Could we make a constant for this value?

parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -100,10 +106,18 @@ func (i *DomainNameOverrideInterceptor) getDomainName() (domainName string, err
...
+		if err != nil {
+			return "", err
+		}
 		return domainName, nil
 	}
 
 	// In any other environment fallback to a generic remote cluster domain
+	_, err = patchCoreDNS(i.kubeClient, `(.*)\.kyma\.example\.com`, i.log)
```
mjakobczyk, line 117: Could we make a constant for this value?

parallel-install/pkg/deployment/coredns.go
mjakobczyk, line 44: Btw - do we need a separate type for this patch? It holds only one value.

parallel-install/pkg/deployment/coredns_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+
+	"github.com/kyma-incubator/hydroform/parallel-install/pkg/logger"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func Test_patchCoreDNS(t *testing.T) {
```
mjakobczyk, line 13: I expect some more tests in-place.

mjakobczyk (changes requested):

parallel-install/pkg/deployment/coredns.go
Tomasz-Smelcerz-SAP: I think the type adds nice context to the value, so it serves the documentation purpose better than a variable name or  a comment. It's OK to have it IMO.

parallel-install/pkg/deployment/overrideinterceptors.go
Tomasz-Smelcerz-SAP: +1, I also don't like package-wide vars, if not necessary.
- We can't make it constant I am afraid (golang limitations)
- It's mutable by design so you never know if a piece of code doesn't change it...

If you need to access the common set of retryOptions from multiple places, I'd just provide a private function that returns a fresh slice on every invocation. This is not "contant" then, but at least any mutations have only local impact then.

parallel-install/pkg/deployment/coredns.go
mjakobczyk: Agree. 👍

werdes72: Done

parallel-install/pkg/deployment/coredns.go
werdes72: Done

parallel-install/pkg/deployment/coredns_test.go
werdes72: DNS has to be uppercase, renamed to `fakeCoreDNSConfigMap`

parallel-install/pkg/deployment/coredns.go
werdes72: We need that in order to fill golang template with some variables, `domainName` in this case. I don't know if there is another way for doing that.

parallel-install/pkg/deployment/coredns_test.go
werdes72: Done

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Done

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Done

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: Done

parallel-install/pkg/deployment/coredns_test.go
```diff
@@ -5,50 +5,83 @@ import (
...
 
-	t.Run("test skipping coreDNS patch", func(t *testing.T) {
+	t.Run("test skipping coreDNS patch when coreDNS deployment doesn't exist", func(t *testing.T) {
 		// given
-		kubeClient := fake.NewSimpleClientset(coreDNSCM)
+		kubeClient := fake.NewSimpleClientset(coreDNSConfigMap)
 
 		// when
-		cm, err := patchCoreDNS(kubeClient, `(.*)\.loacal\.kyma\.dev`, log)
+		cm, err := patchCoreDNS(kubeClient, `(.*)\.local\.kyma\.dev`, log)
```
mjakobczyk, line 24: You repeat the same domain to almost every test case. Maybe it would be worth setting once in before-tests section?

parallel-install/pkg/deployment/coredns_test.go
```diff
@@ -5,50 +5,83 @@ import (
...
+
+		// when
+		cm, err := patchCoreDNS(kubeClient, domain, log)
+
+		// then
+		require.NoError(t, err)
+		require.Contains(t, cm.Data["Corefile"], domain)
+	})
+
+	t.Run("test patching coreDNS configMap when coreDNS configMap is wrong", func(t *testing.T) {
```
mjakobczyk, line 63: What is wrong with the domain? Can you rename the test-case to be more self-descriptive?

parallel-install/pkg/deployment/coredns_test.go
```diff
@@ -5,50 +5,83 @@ import (
...
+		}
+		kubeClient := fake.NewSimpleClientset(coreDNSDeployment, emptyConfigMap)
+		domain := `(.*)\.local\.kyma\.dev`
+
+		// when
+		cm, err := patchCoreDNS(kubeClient, domain, log)
+
+		// then
+		require.NoError(t, err)
+		require.Contains(t, cm.Data["Corefile"], domain)
```
mjakobczyk, line 60: Is this check sufficient? Do we expect `domain` to be anywhere in Corefile, or in a specific place? Is there any way to check it in this shape?

mjakobczyk (changes requested):

parallel-install/pkg/deployment/overrideinterceptors.go
werdes72: I think it will just move code repetition to another place or create more code (another if/switch based on domain name)

parallel-install/pkg/deployment/coredns_test.go
werdes72: I don't see a better way of checking this Corefile quickly, maybe if it's equal to predefined string?

mjakobczyk: If we do not have an easy way of parsing Corefile, then let's leave it this way.

mjakobczyk (approved):
LGTM! 👍

PR #274, opened 2021-04-30
Update go-git and fix commit resolve for semVer.

elchead:
**Description**

- use go-git function to resolve commit hash for different input (branch, tag)
- update go-git

**Related issue(s)**


parallel-install/pkg/git/resolve.go
```diff
@@ -33,30 +32,14 @@ func (rl *remoteRefLister) List(repoURL string) ([]*plumbing.Reference, error) {
 
 var defaultLister refLister = &remoteRefLister{}
 
-// resolveRevision tries to convert a pseudo-revision reference (e.g. semVer, tag, PR, main, etc...) into a revision that can be checked out.
-func ResolveRevision(repo, rev string) (string, error) {
+// ResolvePRrevision tries to convert a PR into a revision that can be checked out.
+func ResolvePRrevision(repoName, rev string) (string, error) {
```
clebs, line 36: I think this functionality should not be exported, it is an internal working of the package.

Therefore having this function is also not bringing any value. I would put this `switch` in `resolveRevision` and make the default case call `go-git`'s function.

parallel-install/pkg/git/clone.go
```diff
@@ -23,46 +25,65 @@ func (rc *remoteRepoCloner) Clone(url, path string, autoCheckout bool) (*git.Rep
 	})
 }
 
-var defaultCloner repoCloner = &remoteRepoCloner{}
+type RepoCloner interface {
```
clebs, line 28: Why introduce this interface and why make it exported? Exported members change the package's public API.

parallel-install/pkg/git/clone.go
```diff
@@ -23,46 +25,65 @@ func (rc *remoteRepoCloner) Clone(url, path string, autoCheckout bool) (*git.Rep
...
+	Clone(dstPath string) error
+	Checkout(rev string) error
+	resolveRevision(rev string) (*plumbing.Hash, error)
+}
 
-// CloneRepo clones the repository in the given URL to the given dstPath and checks out the given revision.
-// revision can be 'main', a release version (e.g. 1.4.1), a commit hash (e.g. 34edf09a) or a PR (e.g. PR-9486).
-func CloneRepo(url, dstPath, rev string) error {
-	rev, err := ResolveRevision(url, rev)
+type Repo struct {
```
clebs, line 34: Why introduce this type and make it exported? Exported members change the package's public API.

parallel-install/pkg/git/clone_test.go
```diff
@@ -62,3 +62,11 @@ func TestCloneRepo(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "Add README\n", commit.Message)
 }
+
+// func TestRealCloneRepo(t *testing.T) {
```
clebs, line 66: Please either uncomment or delete this code.

clebs (changes requested):

parallel-install/pkg/git/clone.go
elchead: I think its nicer to see which methods are expected of the struct. Although there will probably never be another repoCloner the interface makes it more flexible to include another implementation (see factory NewRepo())

parallel-install/pkg/git/clone.go
elchead: I make it private, but my idea was to make the cloneRepo function more readable through encapsulation. Also we might potentially want to have more control over the cloning behavior in the future? The repo class could be a good boundary interface then

parallel-install/pkg/git/clone_test.go
elchead: This is not a unit test, since it actually clones the repo. But I kept it to easily check bugs at the boundary to go-git later (as we had when trying to checkout specific versions). Should I still delete it?

parallel-install/pkg/git/clone.go
clebs: In my opinion, this only makes the package usage more complex and adds more boilerplate without giving any benefits.
It would help thinking about it as package and its usage. It is more simple, readable and idiomatic to do (3 lines vs 7):
```go
if err := git.CloneRepo(url, path, rev); err != nil {
    return err
}
```
instead of:
```go
repo := git.NewRepo(url)
if err := repo.Clone(path); err != nil {
    return err
}
if err := repo.Checkout(rev); err != nil {
    return err
}
```

It is a harder to use API and more error prone. For example, what happens if I call `repo.Checkout()` before `repo.Clone()`? All these wrong API usage problems (the most common usually) go away automatically with a compact and clean API.

Regarding the control, we actually loose some as the previous potential misuse shows. We no longer control the order of the logic for example.

parallel-install/pkg/git/clone.go
clebs: My response to this is more or less the same as for the `Repo` struct. Only makes things harder to use and understand with no benefit.

parallel-install/pkg/git/clone_test.go
clebs: Yes, with modern version control systems leaving commented stuff for casual usage is not necessary.
You can always explore git revisions and copy this snippet when you need it.

parallel-install/pkg/git/resolve_test.go
```diff
@@ -3,8 +3,9 @@ package git
 import (
 	"testing"
 
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/stretchr/testify/require"
-	"gopkg.in/src-d/go-git.v4/plumbing"
+	// "gopkg.in/src-d/go-git.v4/plumbing"
```
clebs, line 8: Please remove this comment

parallel-install/pkg/git/clone.go
```diff
@@ -2,12 +2,28 @@ package git
...
+	if err != nil {
+		return errors.Wrapf(err, "Error downloading repository (%s)", url)
+	}
+	if rev != "" {
+		return checkout(repo, url, rev)
+	}
+	return nil
+}
+
+var defaultCloner repoCloner = &remoteRepoCloner{}
```
clebs, line 25: Please move this to the top. Package variables should always be at the top of the file.

parallel-install/pkg/git/clone.go
```diff
@@ -23,46 +39,32 @@ func (rc *remoteRepoCloner) Clone(url, path string, autoCheckout bool) (*git.Rep
...
-func CloneRepo(url, dstPath, rev string) error {
-	rev, err := ResolveRevision(url, rev)
-	if err != nil {
-		return err
-	}
-
-	if err := CloneRevision(url, dstPath, rev); err != nil {
-		return err
+func resolveRevision(repo *git.Repository, url, rev string) (*plumbing.Hash, error) {
+	err := error(nil)
```
clebs, line 44: This variable declaration is not necessary

clebs (approved):
LGTM!

PR #276, opened 2021-05-06
Add override interceptor for container registry

chrkl:
**Description**

We want to use k3d's container registry for integration tests. This
change adds an override interceptor to set the registry address for the
serverless component to k3d registry.

Using the k3d registry for serverless function builds requires also a
change to the CoreDNS configuration, which will be added in a separate
PR.

**Related issue(s)**

See also https://github.com/kyma-project/cli/issues/671

### DIFF
--- a/parallel-install/pkg/deployment/overrideinterceptors.go
+++ b/parallel-install/pkg/deployment/overrideinterceptors.go
@@ -381,3 +381,90 @@ func defaultRetryOptions() []retry.Option {
 		retry.DelayType(retry.FixedDelay),
 	}
 }
+
+type RegistryDisableInterceptor struct {
+	kubeClient kubernetes.Interface
+}
+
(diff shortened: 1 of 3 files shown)


parallel-install/pkg/deployment/overrideinterceptors.go
```diff
@@ -381,3 +381,57 @@ func defaultRetryOptions() []retry.Option {
...
+		kubeClient:                kubeClient,
+		enableInternalRegistryKey: enableInternalRegistryKey,
+		serverAddressKey:          serverAddressKey,
+		internalServerAddressKey:  internalServerAddressKey,
+		registryAddressKey:        registryAddressKey,
+	}
+}
+
+func (r RegistryInterceptor) String(value interface{}, key string) string {
+	return fmt.Sprintf("%v", value)
```
clebs, line 404: Shouldn't this return the intercepted value instead of the default/user defined in case of k3s?

PR #278, opened 2021-05-10
Add resources labeling

mjakobczyk:
**Description**

- add resources labelling in `PreInstaller

**Related issue(s)**

See https://github.com/kyma-project/hydroform/issues/214


parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -231,3 +242,21 @@ func findOnlyDirectoriesAmong(input []os.FileInfo) (o []os.FileInfo) {
 
 	return o
 }
+
+func updateLabelIfNecessaryOf(obj *unstructured.Unstructured, label string) {
```
clebs, line 246: The name of this function is a bit misleading, in the end it is just adding a given label to the object.
I would just name it `addLabel()`.

werdes72 (approved):
/meow

PR #284, opened 2021-05-13
Expose recoverable attribute in InstallationError

ebensom:
**Description**

Expose recoverable attribute in `InstallationError`. The value is based on whether or not the `action: install` label is present on the installation resource. 
If it's there, the kyma-installer might resolve the error with it's next retry and so the error is still considered as recoverable. If the label is not there, the error is for sure not recoverable as kyma-installer stopped retrying and removed the label.


franpog859 (approved):
LGTM

PR #286, opened 2021-05-18
Introduce CRDs uninstallation

mjakobczyk:
**Description**

- Introduce CRDs uninstallation

**Related issue(s)**

See issue: https://github.com/kyma-project/hydroform/issues/214
See preceding PR: https://github.com/kyma-project/hydroform/pull/283


parallel-install/example/example.go
```diff
@@ -150,6 +145,22 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to uninstall Kyma: %v", err)
 	}
+
+	postUninstallerCfg := postuninstaller.Config{
+		Log:              installationCfg.Log,
+		KubeconfigSource: installationCfg.KubeconfigSource,
```
strekm, line 151: can't `commonRetryOpts` be part of this configuration? then `NewPostUninstaller` would have the same signature , maybe they can even use the same config

parallel-install/pkg/postuninstaller/postuninstaller.go
```diff
@@ -0,0 +1,161 @@
...
+// PostUninstaller removes leftover resources from k8s cluster, added during Kyma installation.
+type PostUninstaller struct {
+	cfg             Config
+	dynamicClient   dynamic.Interface
+	resourceManager preinstaller.ResourceManager
+	retryOptions    []retry.Option
+}
+
+// Output contains lists of Deleted and NotDeleted resources during PostUninstaller uninstallation.
+type Output struct {
```
strekm, line 36: i don't see a reason for having this information tbh

parallel-install/example/example.go
```diff
@@ -150,6 +145,22 @@ func main() {
...
+	}
+
+	postUninstaller, err := postuninstaller.NewPostUninstaller(postUninstallerCfg, commonRetryOpts)
+	if err != nil {
+		log.Fatalf("Failed to create Kyma post-uninstaller: %v", err)
+	}
+
+	uninstallResult, err := postUninstaller.UninstallCRDs()
+	if err != nil || len(uninstallResult.NotDeleted) > 0 {
```
strekm, line 160: is there a possibility that err is nil but still there are some not deleted CDRs or other way around?

parallel-install/pkg/postuninstaller/postuninstaller.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "Error occurred when preparing Kyma CRD label selector")
+	}
+
+	selector = labels.NewSelector()
+	selector = selector.Add(*kymaCrdReq)
+	return selector, nil
+}
+
+func (c *PostUninstaller) fetchCrdsLabeledWith(selector labels.Selector) (crds []unstructured.Unstructured, err error) {
```
strekm, line 96: do we need to grab resources before deleting them? can't we just delete all with given label?

parallel-install/pkg/postuninstaller/postuninstaller.go
```diff
@@ -0,0 +1,161 @@
...
+	}, c.retryOptions...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return
+}
+
+func (c *PostUninstaller) deleteCrdsFrom(crdsList []unstructured.Unstructured) (o Output, err error) {
```
strekm, line 146: try to delete all CRDs using labels

parallel-install/pkg/preinstaller/preinstaller.go
```diff
@@ -34,11 +36,17 @@ import (
 	"k8s.io/client-go/dynamic"
 )
 
+const (
```
strekm, line 39: consider moving this to common file

strekm (changes requested):

parallel-install/pkg/postuninstaller/postuninstaller.go
piotrkpc, line 36: Isn't that code duplication? there is the same public struct in `preinstaller`

parallel-install/pkg/postuninstaller/postuninstaller.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return &PostUninstaller{
+		cfg:             cfg,
+		dynamicClient:   dynamicClient,
+		resourceManager: resourceManager,
+		retryOptions:    retryOptions,
+	}, nil
+}
+
+// UninstallCRDs that belong to Kyma from a k8s cluster.
```
piotrkpc, line 68: I'm personally not a fan of this kind of comments. It does not say much more than the func declaration. I'd love to see comments that answers more to "why? how?" questions.

parallel-install/pkg/postuninstaller/postuninstaller.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (c *PostUninstaller) fetchCrdsLabeledWith(selector labels.Selector) (crds []unstructured.Unstructured, err error) {
+	crdsMap := make(map[string]unstructured.Unstructured)
+	crdGvrV1Beta1 := c.crdGrkWith("v1beta1")
+	crdGvrV1 := c.crdGrkWith("v1")
+	gvrs := [2]schema.GroupVersionResource{crdGvrV1Beta1, crdGvrV1}
+
+	for _, gvr := range gvrs {
+		list, err := c.listResourcesUsing(gvr, selector)
```
piotrkpc, line 103: `resources` is better name than `list` imho

parallel-install/pkg/postuninstaller/postuninstaller_test.go
```diff
@@ -0,0 +1,430 @@
...
+			crd1 := fixCrdResourceWith("crd1", "apiextensions.k8s.io", "v1beta1", "origin", "unknown")
+			crd2 := fixCrdResourceWith("crd2", "apiextensions.k8s.io", "v1beta1", "origin", "unknown")
+			crd3 := fixCrdResourceWith("crd3", "apiextensions.k8s.io", "v1beta1", "origin", "unknown")
+			client := fake.NewSimpleDynamicClientWithCustomListKinds(scheme, fixCrdGvrMap())
+			applyMockObj(t, client, crd1, fixCrdGvrV1Beta1())
+			applyMockObj(t, client, crd2, fixCrdGvrV1Beta1())
+			applyMockObj(t, client, crd3, fixCrdGvrV1Beta1())
+			resourceManager := &mocks.ResourceManager{}
+			uninstaller := getPostUninstaller(cfg, resourceManager, client, retryOptions)
+
```
piotrkpc, line 55: the above makes this tests too large imho, could we refactor/extract this?

piotrkpc:
looks good, I've notice some minor things and tests feels long and could use some refactoring imho but I think you will know better if that's something worth improving at this point.

parallel-install/example/example.go
mjakobczyk: Ultimately I will get rid of the introduced struct and merge it with `Deletion`, so input will no longer be needed.

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: This kind of information was provided by `PreInstaller` too, but now when I change the strategy to bulk delete all labeled resources it would make no sense to fetch the ones that remained because of error, so the struct will be removed. I will consider removing it from `PreInstaller` too.

parallel-install/example/example.go
mjakobczyk: It was not designed as fail-fast approach, so we would always end up with some data in `Output`. If `err` was nil all CRDs should land in either `Installed` or `NotInstalled` state containers.

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: I don't think so. As an improvement I will try to bulk them all. Good point. 👍

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: As above.

parallel-install/pkg/preinstaller/preinstaller.go
mjakobczyk: I will introduce some `constants.go` file.

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: In fact we could reuse it, yes. But I will remove it anyways!

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: I will extend this one with some implementation details of how it does it internally. 👍

parallel-install/pkg/postuninstaller/postuninstaller.go
mjakobczyk: I would even say `listOfResources`, but I got the point!

parallel-install/pkg/postuninstaller/postuninstaller_test.go
mjakobczyk: I will put creation & application into common function wherever applicable (where all resources get the same properties). It should make it slightly shorter then!

parallel-install/pkg/config/constants.go
```diff
@@ -0,0 +1,10 @@
+package config
```
clebs, line 1: Do we really need a separate file for this? I would have expected these in the place where they are most meaningful.

Or at least in `config.go`, but not a dedicated file.

parallel-install/pkg/preinstaller/resourcemanager.go
```diff
@@ -121,14 +129,48 @@ func (c *DefaultResourceManager) UpdateResource(resource *unstructured.Unstructu
...
+	return c.dynamicClient.Resource(gvr).Create(context.TODO(), resource, opts)
+}
+
+func (c *DefaultResourceManager) updateResource(resource *unstructured.Unstructured, gvr schema.GroupVersionResource, opts metav1.UpdateOptions) (*unstructured.Unstructured, error) {
+	return c.dynamicClient.Resource(gvr).Update(context.TODO(), resource, opts)
 }
 
-func (c *DefaultResourceManager) createResource(resource *unstructured.Unstructured, resourceSchema schema.GroupVersionResource) (*unstructured.Unstructured, error) {
-	return c.dynamicClient.Resource(resourceSchema).Update(context.TODO(), resource, metav1.UpdateOptions{})
+func (c *DefaultResourceManager) deleteCollectionOfResources(gvr schema.GroupVersionResource, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error {
```
clebs, line 162: What is the value of having all these unexported functions?

Wouldn't it be the same calling (e.g. in this case) `c.dynamicClient.Resource(gvr).DeleteCollection(...)` directly in the public function?

I fail to see the benefit of having this.

mjakobczyk: In first iteration of `PreInstaller` I introduced these in order to somehow spy on the tested object and mock the function calls (not to call the `dynamicClient` itself at all. Later I discovered the way we can test it in go and use `fake` objects for testing purposes, but these functions were left the same. In fact I can put them to the exported one and we will end up with the same functionality. 👍

parallel-install/pkg/config/constants.go
mjakobczyk: I did not like putting it to `config.go` as these constants are not configurable in any way. But if you think that this place would be more appropriate for them I will move them!

clebs: Well, constants are per definition not configurable. They are just immutable values that are just used in relation to the package.
Same as `http` package would have a `const STATUS_OK = 200`.

clebs (approved):

strekm (approved):
/meow

PR #296, opened 2021-06-01
Make prerequisites fail fast

clebs:
**Description**

- Prerequisite deployment fails as soon as 1 component fails

**Related issue(s)**
https://github.com/kyma-project/hydroform/issues/294

### DIFF
--- a/parallel-install/pkg/deployment/deployment.go
+++ b/parallel-install/pkg/deployment/deployment.go
@@ -156,6 +156,7 @@ InstallLoop:
 					if phase == InstallPreRequisites {
 						err := errors.Wrapf(cmp.Error, "Error deploying prerequisite: %s", cmp.Name)
 						i.processUpdate(phase, ProcessExecutionFailure, err)
+						i.logStatuses(statusMap)
 						return err
 					}
 					errCount++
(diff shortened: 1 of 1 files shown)


parallel-install/pkg/deployment/deployment.go
```diff
@@ -150,6 +152,12 @@ InstallLoop:
 				i.processUpdateComponent(phase, cmp)
 				//Received a status update
 				if cmp.Status == components.StatusError {
+					// prerequisites fail fast
+					if phase == InstallPreRequisites {
+						err := errors.Wrapf(cmp.Error, "Error deploying prerequisite: %s", cmp.Name)
+						i.processUpdate(phase, ProcessExecutionFailure, err)
```
shorim, line 158: maybe it would be also a good idea to log the error `i.cfg.Log.Error(err)`? since we are also doing this in the other cases

clebs: The other cases are calling `i.logStatuses(statusMap)`, I will add that line.

shorim (approved):

strekm (approved):
/meow

PR #307, opened 2021-06-09
Remove all remaining custom resource finalizers

suleymanakbas91:
**Description**

- Remove all remaining custom resource finalizers

**Related issue(s)**

See kyma-project/kyma#11298


rakesh-garimella (approved):

parallel-install/pkg/deployment/deletion.go
```diff
@@ -313,120 +282,53 @@ func (i *Deletion) deleteKymaNamespaces(namespaces []string) error {
...
-			}
-			//remove namespace
-			if err := i.kubeClient.CoreV1().Namespaces().Delete(context.Background(), ns, metav1.DeleteOptions{}); err != nil && !apierr.IsNotFound(err) {
+				i.cfg.Log.Infof("Namespace '%s' still exists with Phase: '%s'", nsT.Name, nsT.Status.Phase)
+				return fmt.Errorf("Namespace '%s' still exists with Phase: '%s'", nsT.Name, nsT.Status.Phase)
+			}, i.retryOptions...)
+			if err != nil {
 				errorCh <- err
 			}
 			i.cfg.Log.Infof("Namespace '%s' is removed", ns)
```
tgorgol, line 334: Nothing critical, but in [line 356](https://github.com/kyma-project/hydroform/pull/307/files#diff-b6c702ed36695a7a97c142d9f909c60d1c11144074eec3c3aa7b9f726570abe9R356) shouldn't we use something like multi error instead of wrapping? Won't the errors from different goroutines get mixed up right now?

parallel-install/pkg/deployment/deletion.go
```diff
@@ -313,120 +282,53 @@ func (i *Deletion) deleteKymaNamespaces(namespaces []string) error {
...
 
-				//HACK: Delete finalizers of leftover ORY Rules
-				ruleResource := schema.GroupVersionResource{
-					Group:    "oathkeeper.ory.sh",
-					Version:  "v1alpha1",
-					Resource: "rules",
-				}
-
-				rules, err := i.dClient.Resource(ruleResource).Namespace(ns).List(context.Background(), metav1.ListOptions{})
+				//HACK: Delete finalizers of leftover Custom Resources
```
tgorgol, line 285: If I understand this correctly we're only removing finalizers on Custom Resources and one specific secret, are we good with not checking core resources for example?

PR #309, opened 2021-06-11
Delete all Custom Resources during uninstallation

suleymanakbas91:
**Description**

- Delete all Custom Resources during uninstallation
- Introduce `--keep-crds` flag
- Fix labeling bug

### DIFF
--- a/parallel-install/pkg/deployment/deletion.go
+++ b/parallel-install/pkg/deployment/deletion.go
@@ -127,7 +127,10 @@ func (i *Deletion) startKymaUninstallation(prerequisitesEng *engine.Engine, comp
 		return err
 	}
 
-	return i.deleteKymaCrds()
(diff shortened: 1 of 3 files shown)


parallel-install/pkg/deployment/deletion.go
```diff
@@ -239,6 +242,68 @@ func (i *Deletion) deleteKymaNamespaces(namespaces []string) error {
 	finishedCh := make(chan bool)
 	errorCh := make(chan error)
 
+	// All the hacks below should be deleted after this issue is done: https://github.com/kyma-project/kyma/issues/11298
```
tobiscr, line 245: would it be possible to move this hack-block into a method so that we can easier separate between the expected code and the workarounds?

tobiscr (approved):

PR #328, opened 2021-07-01
Remove verbose flag param

elchead:
Silencing of deprecation warning is handled by CLI now, so the related code can be removed.

**Description**

- ...
- ...
- ...

**Related issue(s)**
https://github.com/kyma-project/cli/pull/922


elchead:
The verbose parameter needs to be removed in the CLI after merging and a bumped hydroform version.

JeremyHarisch (approved):

PR #334, opened 2021-08-16
Add empty protocolsettings to the subscription CR

pPrecel:
**Description**

- be sure that subscriptions are created with the empty `protocolsettings` field

**Related issue(s)**

https://github.com/kyma-project/cli/issues/938

### DIFF
--- a/function/pkg/resources/unstructured/subscription.go
+++ b/function/pkg/resources/unstructured/subscription.go
@@ -58,9 +58,10 @@ func newSubscriptions(cfg workspace.Cfg, f toUnstructured) ([]unstructured.Unstr
 				Labels:    cfg.Labels,
 			},
 			Spec: types.SubscriptionSpec{
-				Protocol: subscriptionInfo.Protocol,
-				Sink:     sink,
-				Filter:   filter,
+				ProtocolSettings: types.ProtocolSettings{},
+				Protocol:         subscriptionInfo.Protocol,
+				Sink:             sink,
+				Filter:           filter,
 			},
 		}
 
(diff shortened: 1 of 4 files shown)


function/pkg/resources/unstructured/subscription.go
```diff
@@ -58,9 +58,10 @@ func newSubscriptions(cfg workspace.Cfg, f toUnstructured) ([]unstructured.Unstr
 				Labels:    cfg.Labels,
 			},
 			Spec: types.SubscriptionSpec{
-				Protocol: subscriptionInfo.Protocol,
-				Sink:     sink,
-				Filter:   filter,
+				ProtocolSettings: &types.ProtocolSettings{},
```
tgorgol, line 61: The issues says that:
> Subscription CR generated by function Kyma CLI should have protocolsettings property omitted **if it is not overriden by user**

Not sure if I understand that correctly but doesn't that mean that users should be able to override the protocolsettings now?

pPrecel: Nope. It means that the description of the issue is wrong 😃  Because it says that the `protocolsettings` in the Subscription CR (in the cluster) should be omitted (but it can't - see [this](https://kyma-project.io/docs/components/eventing/#custom-resource-subscription-custom-resource-parameters)) and the user should not set this property (we don't support this functionality and we don't want to)

In my previous contribution, I set the `protocolsettings` to be always omitted and it, unfortunately, doesn't work  ([here](https://github.com/kyma-project/hydroform/pull/331/files))

tgorgol (approved):

PR #352, opened 2022-03-31
Fix for local functions running from subfolders

Cortey:
**Description**
Fix that allows user to move their function resources to subfolders and still run the function locally using `kyma run function`
**See also**
https://github.com/kyma-project/cli/pull/1231
https://github.com/kyma-project/cli/issues/1122


function/pkg/docker/runtimes/runtimes.go
```diff
@@ -83,6 +88,37 @@ func ContainerCommands(runtime types.Runtime, debug bool, hotDeploy bool) []stri
...
+				Type:   mount.TypeBind,
+				Source: workDir,
+				Target: KubelessPath,
+			},
+		}
+	}
+}
+
+func MoveInlineCommand(sourcePath, depsPath string) []string {
+	return []string{
```
moelsayed, line 116: nit: can we make this a bit more readable? you can put the output of join in variables and then pass them to Sprintf. I think it would be easier to read and understand.

moelsayed, line 116: nit: can we make this a bit more readable? you can put the output of join in variables and then pass them to Sprintf. I think it would be easier to read and understand.

moelsayed (commented):
Looks food. One nit comment.

moelsayed (approved):

PR #362, opened 2022-05-11
Move labels from spec to metadata in function

dbadura:
**Description**

- use `metadata.labels` instead of `spec.labels`

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/11155


function/pkg/workspace/synchronise_test.go
```diff
@@ -397,6 +397,9 @@ func Test_Synchronise(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			//TODO: Please refactor this test because:
```
pPrecel, line 400: 👍

Issue #368, opened 2022-06-14
Function is unchanged when label or annotation change.


dbadura:
**Description**
When user add label or annotation to function, the sdk says that the object is the same, but it's not.

**Expected result**
When function is applied with changed label or annotations, the hydroform updates the labels or annotations.

**Actual result**
Currently the labels and annotations are not detected as change.
The faulty code is here: https://github.com/kyma-project/hydroform/blob/main/function/pkg/operator/operator.go#L41=

**Steps to reproduce**
1. Use kyma cli and init basic function: `kyma function init`
2. Apply function: `kyma function apply .`
3. Add any label to config.yaml, for example:
```
labels:
  aa: bb
```
4. apply changes: `kyma function apply`

The cli will report that the function didn't change, but it's not true.

**Troubleshooting**

PR #369, opened 2022-06-21
Update function on label or annotation change 368

anoipm:
**Description**

- Detection whether labels and annotations was changed was added while applying the function.
- Updating labels and annotations was added while applying the function.
- Change detection worked only in one direction (detects added elements but not removed - equality.Semantic.DeepDerivative vs equality.Semantic.DeepEqual) - fixed.

**Related issue(s)**
See also #368

### DIFF
--- a/function/pkg/operator/operator.go
+++ b/function/pkg/operator/operator.go
@@ -76,13 +76,8 @@ func applyObject(ctx context.Context, c client.Client, u unstructured.Unstructur
 	return response, statusEntryCreated, nil
 }
 
-// updateConfigurationObject copies some elements from the source configuration to the destination configuration.
-// The following items are copied:
-//  * spec,
-//  * metadata/labels,
-//  * metadata/annotations.
-// The rest is unchanged.
 func updateConfigurationObject(destination *unstructured.Unstructured, source unstructured.Unstructured) *unstructured.Unstructured {
+	// Copies elements: `spec`, `metadata/labels`, `metadata/annotations` from source to destination. The rest is unchanged.
 	destination.Object["spec"] = source.Object["spec"]
 
 	destinationMetadata := getMetadataFromConfigurationObject(*destination)
@@ -103,12 +98,8 @@ func updateConfigurationElement(destination *map[string]interface{}, source map[
 	}
 }
 
-// configurationObjectsAreEquivalent compares two structures that represent configuration. Checks that the elements:
-//  * spec,
-//  * metadata/labels,
-//  * metadata/annotations
-// are equal (semantically). The rest are ignored.
 func configurationObjectsAreEquivalent(first unstructured.Unstructured, second unstructured.Unstructured) bool {
+	// Checks that the elements: `spec`, `metadata/labels`, `metadata/annotations` are equal (semantically).
 	specAreEqual := equality.Semantic.DeepEqual(first.Object["spec"], second.Object["spec"])
 
 	firstMetadata := getMetadataFromConfigurationObject(first)
--- a/function/pkg/operator/operator.go
+++ b/function/pkg/operator/operator.go
@@ -48,8 +48,6 @@ func applyObject(ctx context.Context, c client.Client, u unstructured.Unstructur
 	// If object needs update
 	if objFound && !equal {
 		response = updateConfigurationObject(response, u)
-		//response.Object["spec"] = u.Object["spec"]
-		//TODO: add copy of metadata/labels and metadata/annotations
 		err = retry.RetryOnConflict(retry.DefaultRetry, func() (err error) {
 			response, err = c.Update(ctx, response, metav1.UpdateOptions{
 				DryRun: stages,
--- a/function/pkg/operator/operator.go
+++ b/function/pkg/operator/operator.go
@@ -47,7 +47,8 @@ func applyObject(ctx context.Context, c client.Client, u unstructured.Unstructur
 
 	// If object needs update
 	if objFound && !equal {
-		response.Object["spec"] = u.Object["spec"]
+		response = updateConfigurationObject(response, u)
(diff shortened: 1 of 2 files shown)


function/pkg/operator/operator.go
```diff
@@ -48,7 +47,9 @@ func applyObject(ctx context.Context, c client.Client, u unstructured.Unstructur
 
 	// If object needs update
 	if objFound && !equal {
-		response.Object["spec"] = u.Object["spec"]
+		response = updateConfigurationObject(response, u)
+		//response.Object["spec"] = u.Object["spec"]
+		//TODO: add copy of metadata/labels and metadata/annotations
```
pPrecel, line 52: During the offline conversation, we figured out that these comments are not valid anymore

function/pkg/operator/operator.go
```diff
@@ -77,6 +78,77 @@ func applyObject(ctx context.Context, c client.Client, u unstructured.Unstructur
...
+	} else {
+		delete(*destination, elementName)
+	}
+}
+
+// configurationObjectsAreEquivalent compares two structures that represent configuration. Checks that the elements:
+//  * spec,
+//  * metadata/labels,
+//  * metadata/annotations
+// are equal (semantically). The rest are ignored.
```
pPrecel, line 112: Let's discuss it if we want to add comments for private methods. Maybe a better option would be to add a comment inside the method (if it's not self-explainable)? But I understood this function after a quick look at it :D

Adding comments for methods is a good practice, I can agree. But if I'm not wrong it is valid for public methods which can be imported by other components/services. An example [here](https://github.com/kubernetes/kubectl/blob/4483ba9cf6c54db2163399ba6276173f22cce75f/pkg/cmd/apply/apply.go#L231)

function/pkg/operator/operator_test.go
```diff
@@ -236,6 +251,31 @@ func Test_applyObject(t *testing.T) {
...
+					result := mockclient.NewMockClient(ctrl)
+
+					result.EXPECT().
+						Get(gomock.Any(), gomock.Any(), gomock.Any()).
+						Return(testObj.DeepCopy(), nil).
+						Times(1)
+
+					result.EXPECT().
+						Update(gomock.Any(), gomock.Any(), gomock.Any()).
+						Return(testObjWithLabelsAndAnnotations.DeepCopy(), nil).
```
pPrecel, line 267: A good option would be to not receive `gomock.Any()` and to not return copied structure. We can add validation of the given argument by replacing this `Any` method with the desired struct

function/pkg/operator/operator_test.go
```diff
@@ -483,6 +519,368 @@ func Test_waitForObject(t *testing.T) {
...
+				first:  defaultConfigurationObject,
+				second: defaultConfigurationObject,
+			},
+			want: true,
+		},
+		{
+			name: "similar configuration objects - changed metadata/^(labels|annotations)/*",
+			args: args{
+				first:  defaultConfigurationObject,
+				second: similarConfigurationObjectWithChangedOmmitedPartOfMetadata,
```
pPrecel, line 608: I see here possible two small improvements. 
1. Name of this field is definitely too long :D To be honest if the longest name I've ever seen
2. Instead of doing this operation above test tables you can use the `func() any_tyle { ... }()` anonymous method to keep this logic close to a specific test case

function/pkg/operator/operator_test.go
```diff
@@ -244,12 +284,8 @@ func Test_applyObject(t *testing.T) {
...
 				return
 			}
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("applyObject() got = %v, want %v", got, tt.want)
-			}
-			if !reflect.DeepEqual(got1, tt.want1) {
-				t.Errorf("applyObject() got1 = %v, want %v", got1, tt.want1)
-			}
+			require.Equal(t, tt.want, got)
+			require.Equal(t, tt.want1, got1)
```
pPrecel, line 288: Good job 👍

pPrecel (commented):
good job! I see a few possible improvements only 🥇

function/pkg/operator/operator_test.go
pPrecel: regarding this first point, I can share with you two good materials about code style and good practices in go:
1. https://go.dev/doc/effective_go <- official one
2. https://github.com/uber-go/guide/blob/master/style.md <- I really like this one

function/pkg/operator/operator.go
anoipm: Removed

function/pkg/operator/operator_test.go
anoipm: ad 1 (from first comment): I have no idea how to change these names without losing their meaning. I prefer long names than `testObj` and` testObj1`.

ad 2: All configuration objects prepared for tests that are not reused have been moved to built-in functions.

function/pkg/operator/operator_test.go
anoipm: Fixed

function/pkg/operator/operator_test.go
anoipm: When I have variable `invoice` I know that is something connected with invoice. When I see `testObj` I know only that it's something for tests. 

IMO when I have `destinationConfigurationObjectWithoutLabelsAndAnnotations` I know that it is specific object described in name and I see it in every place where I use it (not only in place with comment). Now we have no problem with too small screens and may use long names. I prefer it but I'm curious arguments against it.

anoipm: We talked about this and decided that I will remove these long named objects because reusing them creates a coupling between different tests.

anoipm: Done

m00g3n (approved):
LGTM

PR #456, opened 2024-06-18
Remove sec-scanners-config.yaml

Cortey:
**Description**
Deleting since the first CLI without the use of hydroform got released

**Related issue(s)**


halamix2 (approved):
/pony

